PR #5, opened 2026-04-01
Add errdefer fix, heap fallback, ERT migration, and theme sync

dakra:
## Summary

- **Add missing errdefer for mouse_encoder** — consistent cleanup guard in `Terminal.init()`, matching all other resources
- **Add heap fallback for HTML formatter buffer** — falls back to 1 MiB heap allocation when the 256KB stack buffer is insufficient for hyperlink scanning on very large terminals
- **Migrate test suite to ERT** — replace custom assertion framework with Emacs's built-in ERT library for better diagnostics, selective test running, and IDE integration (net -209 lines)
- **Add automatic theme color synchronization** — `ghostel-sync-theme` re-applies the color palette to all active terminals when the Emacs theme changes, via `enable-theme-functions` (Emacs 29+) or `load-theme` advice

## Test plan

- [ ] `make test` passes (pure Elisp ERT tests)
- [ ] `make byte-compile` is warning-free
- [ ] CI byte-compile passes across Emacs 27.2, 29.4, snapshot
- [ ] CI native build succeeds on Linux and macOS
- [ ] Manually verify theme sync: `M-x load-theme` updates terminal colors in a running ghostel buffer

### DIFF
--- a/test/ghostel-test.el
+++ b/test/ghostel-test.el
@@ -784,6 +784,42 @@
       (kill-buffer buf))))
 
 ;; -----------------------------------------------------------------------
+;; Test: theme synchronization
+;; -----------------------------------------------------------------------
+
+(ert-deftest ghostel-test-sync-theme ()
+  "Test that ghostel-sync-theme applies palette and redraws ghostel buffers."
+  (let ((palette-calls nil)
+        (redraw-calls nil))
+    (cl-letf (((symbol-function 'ghostel--apply-palette)
+               (lambda (term) (push term palette-calls)))
+              ((symbol-function 'ghostel--redraw)
+               (lambda (term) (push term redraw-calls))))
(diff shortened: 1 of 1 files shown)


dakra:
**Re: heap buffer caching (render.zig)** — The heap path is a rare fallback that only triggers when the 256KB stack buffer is insufficient, which requires an unusually large terminal with many hyperlinks. Caching a 1 MiB buffer in the Terminal struct would add complexity for a path that almost never executes. The page allocator is efficient for this size (mmap/munmap), so the overhead is negligible in the rare case it's needed.

**Re: theme sync test** — Good call. Added in 3e3db6a — the new `ghostel-test-sync-theme` test stubs `ghostel--apply-palette` and `ghostel--redraw`, then verifies palette is applied to ghostel-mode buffers, redraw is triggered, and redraw is skipped when copy mode is active.

PR #54, opened 2026-04-07
Add evil-mode integration

### LINKED ISSUE #52: Support evil movement
Hi, thanks for the project, really excited about it! I use `evil-mode` and have noticed an issue. To reproduce

1. turn on `evil-mode`
2. enter `insert` mode and enter a few character
3. enter `normal` mode and move your cursor back 1 character
4. type one character
5. expect that one character to be inserted at cursor, but instead it is inserted at the end of the string you typed in step 2 above

Here's a demo in case visuals help

https://github.com/user-attachments/assets/640c5777-1379-4182-b17c-7add72a3971f

dakra:
## Summary
- Adds `ghostel-evil.el` with cursor synchronization between Emacs point and terminal cursor during evil state transitions
- Entering normal state snaps point to the terminal cursor; entering insert state (`i`/`a`/`I`/`A`) sends arrow keys to move the terminal cursor to point
- Alt-screen programs (vim, less) are unaffected
- Adds evil-mode section to README
- Splits melpazoid CI into two jobs: one for ghostel, one for ghostel-evil (with evil installed)

Closes #52

## Test plan
- [x] Open ghostel with evil-mode enabled
- [x] Type text in insert mode, ESC to normal, move with hjkl, press `i` — characters should appear at cursor position
- [x] Verify `a`, `I`, `A` also insert at the correct position
- [x] Verify alt-screen apps (vim, less) are unaffected
- [x] Byte-compile: `emacs --batch -Q -L . -f batch-byte-compile ghostel-evil.el`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,31 @@ jobs:
             -l test/ghostel-test.el \
             -f ghostel-test-run-elisp
 
+  test-evil:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        emacs_version:
+          - '29.4'
+          - 'snapshot'
+    steps:
+      - uses: actions/checkout@v4
+      - uses: purcell/setup-emacs@master
+        with:
+          version: ${{ matrix.emacs_version }}
+
+      - name: Install evil
+        run: |
+          git clone --depth 1 https://github.com/emacs-evil/evil.git /tmp/evil
+
+      - name: Run ghostel-evil tests
+        run: |
+          emacs --batch -Q -L /tmp/evil -L . \
+            -l ert \
+            -l test/ghostel-evil-test.el \
+            -f ghostel-evil-test-run-elisp
+
   build-native:
     strategy:
       fail-fast: false
--- a/.github/workflows/melpazoid.yml
+++ b/.github/workflows/melpazoid.yml
@@ -19,7 +19,34 @@ jobs:
       - name: Run
         env:
           LOCAL_REPO: ${{ github.workspace }}
-          RECIPE: (ghostel :fetcher github :repo "dakra/ghostel")
+          RECIPE: (ghostel :fetcher github :repo "dakra/ghostel" :files ("ghostel.el" "ghostel-debug.el" "ghostel-module.*"))
+          EXIST_OK: false
+          WARN_IS_ERROR: true
+        run: make -C ~/melpazoid
+
+  build-evil:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with: { python-version: '3.10' }
+      - name: Install
+        run: |
+          sudo apt-get install emacs && emacs --version
+          git clone https://github.com/riscy/melpazoid.git ~/melpazoid
+      - name: Install dependencies
+        run: |
+          emacs --batch -Q \
+            --eval "(require 'package)" \
+            --eval "(add-to-list 'package-archives '(\"melpa\" . \"https://melpa.org/packages/\") t)" \
+            --eval "(package-initialize)" \
+            --eval "(package-refresh-contents)" \
+            --eval "(package-install 'evil)"
+      - name: Run
+        env:
+          LOCAL_REPO: ${{ github.workspace }}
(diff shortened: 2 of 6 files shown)


curtainp:
Now, `a` works, but `I` and `A` still not work(both insert at current position like `i`). BTW: with [8132da3](https://github.com/dakra/ghostel/commit/8132da37a62bbc14211c7271177bb2d830850fe4)

> @curtainp Thanks for testing! The bugs with `a`, `I`, and `A` should be fixed with the latest update in [5f70af9](https://github.com/dakra/ghostel/commit/5f70af9bd96e54e8239ff4319680dd5eb164838e)
> 
> Can you try again and report back?
> 
> Regarding `D` - that's a deeper limitation since evil deletes text in the Emacs buffer but the terminal doesn't know about it. Supporting edit operators like `d`, `c`, `x` etc. would require mapping them to terminal editing commands (e.g. Ctrl-K for kill-to-end-of-line), which is a bigger piece of work.

dakra:
@curtainp  Thank you so much for testing.

Now I had a closer look at how the vterm implementation works.

Can you test with the latest changes? This time also `d`, `c` etc are implemented.
I had a quick test locally and there everything seemed to work.

curtainp:
Yeah, It still weird. example: `cw` and `dw` will replace current word with a strange symbol like `~`, I don't even know where that symbol came from.
image

dakra:
@curtainp This happened because your shell doesn't interpret the `<delete>` key, and mine does. That's why it worked for me.
Now I use `<backspace>`, which should work everywhere.
Can you try again?

Thanks for your testing-support ❤️

curtainp:
Woo, now `c` and `d` operator works! the only drawback is that the cursor position tends to drift when `d{motion}`. But anyway, thanks for your work! Hope to make progress for others too.(`I` and `A` and cursor position).

https://github.com/user-attachments/assets/b5a580d7-1ecb-4b1b-b1ff-777e643141d1

dakra:
> the only drawback is that the cursor position tends to drift when d{motion}

can you explain it a bit more? in the video of yours, it looks like everything is working? but I it doesn't show what exactly you types, so not sure.

> Hope to make progress for others too.(I and A and cursor position).

Does `I` and `A` not work? it basically just send the `home` and `end` key to the shell.

curtainp:
> can you explain it a bit more? in the video of yours, it looks like everything is working? but I it doesn't show what exactly you types, so not sure.

you see 6s in video, i `dw` last word and expect that cursor should right next to the last single quote, but it does not.


`I` and `A` just as before(insert at current position) with my zsh shell.

jasonatwell:
It's working pretty well for me, the only error I still see is that I can delete everything in the buffer using `ctrl+w` but it's restored as soon as I insert a character. See recording. `ghostel` is now usable for me, that error isn't a blocker, thank you so much for the quick turnaround on this!




https://github.com/user-attachments/assets/7341883c-1546-425b-ba2d-e340f426d31d

dakra:
> I and A just as before(insert at current position) with my zsh shell.

Thanks. Looks like `home` and `end` keys are not bound by default in zsh.
Now I send `C-a` and `C-e` which in my brief testing seems to work everywhere.

Can you check again? Hopefully the last time 🤞

dakra:
> the only error I still see is that I can delete everything in the buffer using ctrl+w but it's restored as soon as I insert a character

this should be fixed now. can you try again?
(before C-w was an evil-command that modified the buffer directly, now I just send C-w straight to the terminal)

jasonatwell:
`ctrl-w` now works as expected. I noticed two other errors
1. `ctrl-u` works like `ctrl-w` was working earlier, it seems to delete the entire line but then inserting a character restores the line
2. using `k` and `j` in `normal-mode` to navigate, then entering `insert` mode doesn't work as expected. To reproduce
    1. use `k` in `normal-mode` to move the cursor up
    2. enter `insert` mode
    3. the prompt behaves as though you pressed `up` to cycle to the previous command. If you use `k` two times in step i, it will cycle twice, etc

dakra:
@jasonatwell I added a bunch more ctrl-.. combos in insert mode. a few where still missing.
the up/down error should also be fixed now. when navigating up/down in normal mode, it'll just snap back to the prompt line when entering insert.

curtainp:
@dakra Woo! that's fantastic！ Almost operator work as expected now. could merge it?

dakra:
Merged into main.
Thank you very much for all the testing @curtainp @jasonatwell 

If you find any other bugs, don't hesitate to open a new issue.

PR #61, opened 2026-04-08
Add Kitty graphics protocol support

dakra:
## Summary

- Intercept Kitty graphics APC_G escape sequences in the Zig native module, parsing transmit/display/delete commands with chunked transfer and base64 payload support
- Store decoded images as PNG (passthrough) or PPM (converted from RGB/RGBA pixel data) in a per-terminal image store
- Track cursor position at each placement command to determine image location in the viewport
- Render images in Emacs buffers using sliced `display` text properties so multi-row images fit the fixed terminal grid
- Pass cell pixel dimensions (`frame-char-width` / `frame-char-height`) during resize for correct image sizing and libghostty size reports

## Limitations (follow-up work)

- Zlib-compressed payloads (`o=z`) not yet supported
- Unicode virtual placeholders (U+10EEEE) not yet rendered
- Images don't survive scrollback (viewport-relative placement tracking)
- File/shared-memory transmission mediums not supported (direct only)

## Test plan

- [x] `zig build check` — compiles cleanly
- [x] `./build.sh` — full build with linking succeeds
- [x] Elisp test suite — all 38 tests pass
- [x] Byte-compile — no warnings
- [x] Manual: `timg --pixelation=kitty <image>` in a ghostel terminal (GUI Emacs)
- [x] Manual: verify graceful degradation in terminal Emacs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,19 @@ jobs:
       - name: Run package-lint, checkdoc, docquotes
         run: make package-lint checkdoc docquotes
 
+  test-zig:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Zig
+        uses: mlugg/setup-zig@v2
+        with:
+          version: 0.15.2
+
+      - name: Run pure-Zig unit tests
+        run: zig build test
+
   test:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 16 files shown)


dakra:
@emil-e do you want to have a quick look at this PR?

I was using it the last few days, but honestly I don't have much use for graphics inside the terminal.
All my tests with `timg` or `yazi` work and look good.

It adds a lot of code but 8k lines is the vendored in stb png decoder.
And the rest is very independent from our existing code, so I feel it is pretty safe and worst case, easy to revert.

emil-e:
@dakra I'd be happy to take a look when I have some time!

dakra:
> @dakra I'd be happy to take a look when I have some time!

no rush.
I merged it now as a few people already asked for kitty graphics and I think it doesn't harm the
normal user even if the kitty graphics implementation still has some bugs.
But my tests with `yazi` work nicely now.

PR #74, opened 2026-04-10
Fix: wrong number of argument for `ghostel-evil--around-delete` when calling `evil-delete-char`

junyi-hou:
After loading `ghostel-evil`, calling `evil-delete-char` will result in error
```
apply: Wrong number of arguments: #[(orig-fn beg end type register yank-handler) ((if (ghostel-evil--active-p) (progn (if register nil (let ((text (filter-buffer-substring beg end))) (if (string-match-p "
" text) nil (evil-set-register 45 text)))) (let ((evil-was-yanked-without-register nil)) (evil-yank beg end type register yank-handler)) (if (eq type 'line) (progn (ghostel--send-encoded "e" "ctrl") (ghostel--send-encoded "u" "ctrl")) (ghostel-evil--delete-region beg end))) (funcall orig-fn beg end type register yank-handler))) (t) nil "Intercept `evil-delete' in ghostel buffers.
Yanks text to REGISTER, then deletes via PTY.
Covers d, dd, D, x, X."], 5
```
This is due to the fact that `evil-delete-char` is called with only 4 arguments `(evil-delete-char BEG END &optional TYPE REGISTER)`.

This PR fix this by setting the fifth argument, `yank-handler` as optional in the `ghostel-evil--around-delete` advice.

### DIFF
--- a/ghostel-evil.el
+++ b/ghostel-evil.el
@@ -169,7 +169,8 @@ escape sequence is not bound in all shell configurations."
 ;; Advice for evil editing operators
 ;; ---------------------------------------------------------------------------
 
-(defun ghostel-evil--around-delete (orig-fn beg end type register yank-handler)
+(defun ghostel-evil--around-delete
+    (orig-fn beg end &optional type register yank-handler)
   "Intercept `evil-delete' in ghostel buffers.
 Yanks text to REGISTER, then deletes via PTY.
 Covers d, dd, D, x, X."
--- a/test/ghostel-evil-test.el
+++ b/test/ghostel-evil-test.el
@@ -351,6 +351,24 @@ Uses mocks for native functions."
        (should (cl-find '("u" . "ctrl") keys-sent :test #'equal))
        (should (cl-find '("e" . "ctrl") keys-sent :test #'equal))))))
 
+(ert-deftest ghostel-evil-test-delete-char ()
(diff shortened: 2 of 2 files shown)


dakra:
Fix LGTM. One suggestion before merging: please add a regression test so this arity mismatch doesn't silently come back. There's already a precedent at [`test/ghostel-evil-test.el:377`](https://github.com/dakra/ghostel/blob/main/test/ghostel-evil-test.el#L377) (`ghostel-evil-test-change-whole-line`) guarding the equivalent `delete-func` bug in `ghostel-evil--around-change`.

I've prepared one on the `claude/review-pr-74-qy1R2` branch in this repo — commit [`d196188`](https://github.com/dakra/ghostel/commit/d196188). Feel free to cherry-pick it into this PR:

```sh
git fetch origin claude/review-pr-74-qy1R2
git cherry-pick d196188
```

Also, once the test is in, please squash the two commits into a single one before merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

junyi-hou:
updated with:
- cherry-picked the test from `claude/review-pr-74-qy1R2`
- adopt copilot review (use `&rest` and `apply`)

CI run successful in the forked repo: https://github.com/junyi-hou/ghostel/actions/runs/24262932913

dakra:
Merged. Thank you very much ❤️

PR #78, opened 2026-04-11
Add eat-style input modes (char, emacs, copy, line, semi-char)

### LINKED ISSUE #40: Project plans in relation to eat.el-like input modes
Hello,

Thanks for your work on the project.

I have been a long time user of [EAT](https://elpa.nongnu.org/nongnu/eat.html) but plan to keep an eye on this project and potentially switch to it in the future depending on how maintained it is and the features it grows to have.

One question I have is related to the trajectory of the feature set, namely, whether it plans to implement input modes similar to the one EAT has. (You can read more about the EAT input modes [here](https://elpa.nongnu.org/nongnu/doc/eat.html#Input-Modes).) Currently there is a "copy" input mode but I would like clarification on whether that mode is the only one that is intended to be fleshed out or whether in the future (just for example) the "copy" mode may become like the EAT "emacs mode" and an equivalent for the EAT "line mode", "char mode," and "semi-char" mode will emerge.

dakra:
## Summary

Adds five eat.el-inspired input modes for ghostel. Closes #40.

- **semi-char** (default): terminal input, `C-c` prefix reserved
- **char** (`C-c M-d`): all keys to terminal, `M-RET` exits — for TUI apps that want `C-x`, `M-x`, `C-h`
- **Emacs** (`C-c C-e`): **unfrozen** — terminal keeps running, buffer is read-only, standard Emacs keys (`isearch`, `occur`, `M-x`, `C-SPC`+`M-w`) fall through to the global map. Lets you search across a live log
- **copy** (`C-c C-t`): frozen + read-only + aggressive copy keymap (`M-w` copies and exits)
- **line** (`C-c C-l`): buffers input locally in Emacs; `RET` sends the whole line to the shell in one write. Full Emacs editing (yank, transpose, kill-word…) works. Requires OSC 133 shell integration

## Mechanics

- Single state variable `ghostel--input-mode`; `ghostel--copy-mode-active` removed
- Keymap split: slim base `ghostel-mode-map` (C-c prefix + drag-n-drop) with per-mode child maps
- New predicates `ghostel--buffer-editable-p`, `ghostel--terminal-live-p`, `ghostel--terminal-frozen-p` — all five check sites that used `copy-mode-active` migrated
- Emacs mode reuses the existing `inhibit-read-only`/`inhibit-redisplay` bindings around `ghostel--redraw`, so nothing in the Zig layer changes; the delayed-redraw path now preserves point unconditionally in Emacs mode
- Line mode suppresses redraws while composing and forces a full catch-up redraw on exit — no save/restore-around-redraw race
- Prompt nav (`C-c C-n`/`C-c C-p`) now auto-enters Emacs mode instead of copy mode so the terminal keeps running during navigation

## Base

This PR targets `scrollback-in-buffer` (PR #73) since it depends on the always-materialized-scrollback model. Once #73 merges to main, this can be rebased onto main.

## Test plan

- [x] `make build` — native module compiles (no Zig changes, sanity check)
- [x] `make test-all` — 120 ghostel tests pass (93 existing + 27 new mode tests)
- [x] `make test-evil` — 30 evil-integration tests still pass
- [x] `make lint` — package-lint + checkdoc clean
- [x] Manual smoke test: `C-c M-d` char mode, `C-c C-e` Emacs mode + isearch across streaming output, `C-c C-t` copy mode, `C-c C-l` line mode with bash + `ls -la` + RET
- [x] CI green (all matrix jobs)

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,36 @@ string match -qr '^ghostel(,|$)' -- "$INSIDE_EMACS"; and source "$EMACS_GHOSTEL_
 ```
 </details>
 
-## Key Bindings
+## Input modes
 
-### Terminal mode
+Ghostel offers five eat.el-style input modes.  You enter a ghostel
+buffer in **semi-char mode**; switch modes with the key bindings below
+and watch `mode-line-process` for the current mode indicator.
+
+| Mode        | Indicator | Terminal | Buffer      | Purpose                                       |
+|-------------|-----------|----------|-------------|-----------------------------------------------|
+| semi-char   | *(none)*  | live     | editable    | default — type to terminal, `C-c` reserved    |
+| char        | `:Char`   | live     | editable    | TUI apps — *all* keys go to the terminal      |
+| Emacs       | `:Emacs`  | live     | read-only   | search/read while the terminal keeps running  |
+| copy        | `:Copy`   | frozen   | read-only   | precise text selection without scroll churn   |
+| line        | `:Line`   | live     | editable    | compose input with Emacs keys, send on `RET`  |
+
+### Mode-switch keybindings (available from every live mode)
+
+| Key       | Action                                    |
+|-----------|-------------------------------------------|
+| `C-c C-j` | Switch to semi-char mode (universal exit) |
+| `C-c M-d` | Switch to char mode                       |
+| `C-c C-e` | Switch to Emacs mode                      |
+| `C-c C-t` | Toggle copy mode                          |
+| `C-c C-l` | Switch to line mode                       |
+| `M-RET`   | Char mode only: return to semi-char       |
+
+### Semi-char mode (default)
+
+Most keys are sent to the terminal.  Keys in
+`ghostel-keymap-exceptions` (default: `C-c`, `C-x`, `C-u`, `C-h`,
+`C-g`, `M-x`, `M-o`, `M-:`, `C-\`) pass through to Emacs.
 
 | Key         | Action                                 |
 |-------------|----------------------------------------|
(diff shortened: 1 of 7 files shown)


krisbalintona:
FYI I'm trying out this branch and calling `ghostel-download-module` causes my Emacs to crash. But calling Emacs right after shows no apparent breakage. Example:
```
Fatal error 11: Segmentation fault
Backtrace:
emacs(+0x1c0fcc) [0x6391290effcc]
emacs(+0x529e8) [0x639128f819e8]
emacs(+0x52f39) [0x639128f81f39]
emacs(+0x52f40) [0x639128f81f40]
emacs(+0x1bf6cc) [0x6391290ee6cc]
/usr/lib/libc.so.6(+0x3e2d0) [0x7759b0cc22d0]
/lib64/ld-linux-x86-64.so.2(+0xabe3) [0x7759b5a24be3]
/lib64/ld-linux-x86-64.so.2(+0xb0fc) [0x7759b5a250fc]
/lib64/ld-linux-x86-64.so.2(+0xba1c) [0x7759b5a25a1c]
/usr/lib/libc.so.6(+0x168e3e) [0x7759b0dece3e]
/usr/lib/libc.so.6(+0x93010) [0x7759b0d17010]
/lib64/ld-linux-x86-64.so.2(_dl_catch_exception+0xa6) [0x7759b5a1c456]
/lib64/ld-linux-x86-64.so.2(+0x25a9) [0x7759b5a1c5a9]
/usr/lib/libc.so.6(+0x92a03) [0x7759b0d16a03]
/usr/lib/libc.so.6(dlsym+0x8e) [0x7759b0d170ae]
emacs(+0x26fa54) [0x63912919ea54]
/home/krisbalintona/.emacs.d/var/eln-cache/31.0.50-ef4348d8/ghostel-d6cb9d6f-e46e862b.eln(F67686f7374656c2d646f776e6c6f61642d6d6f64756c65_ghostel_download_module_0+0x141) [0x77593a8efac1]
emacs(+0x23a6b4) [0x6391291696b4]
emacs(+0x23235b) [0x63912916135b]
emacs(+0x23a6b4) [0x6391291696b4]
emacs(+0x232f60) [0x639129161f60]
/home/krisbalintona/emacs/31.0/bin/../lib/emacs/31.0.50/native-lisp/31.0.50-ef4348d8/preloaded/simple-fab5b0cf-a4ed70fb.eln(F636f6d6d616e642d65786563757465_command_execute_0+0x2a5) [0x7759abb540d5]
emacs(+0x23a6b4) [0x6391291696b4]
/home/krisbalintona/emacs/31.0/bin/../lib/emacs/31.0.50/native-lisp/31.0.50-ef4348d8/preloaded/simple-fab5b0cf-a4ed70fb.eln(F657865637574652d657874656e6465642d636f6d6d616e64_execute_extended_command_0+0x1e1) [0x7759abb52e91]
emacs(+0x23a6b4) [0x6391291696b4]
emacs(+0x23235b) [0x63912916135b]
emacs(+0x23a6b4) [0x6391291696b4]
emacs(+0x23aa80) [0x639129169a80]
emacs(+0x233deb) [0x639129162deb]
/home/krisbalintona/emacs/31.0/bin/../lib/emacs/31.0.50/native-lisp/31.0.50-ef4348d8/preloaded/simple-fab5b0cf-a4ed70fb.eln(F636f6d6d616e642d65786563757465_command_execute_0+0x2a5) [0x7759abb540d5]
emacs(+0x23a6b4) [0x6391291696b4]
emacs(+0x1b2db8) [0x6391290e1db8]
emacs(+0x2353b4) [0x6391291643b4]
emacs(+0x19cab6) [0x6391290cbab6]
emacs(+0x23530e) [0x63912916430e]
emacs(+0x19ca53) [0x6391290cba53]
emacs(+0x1a5242) [0x6391290d4242]
emacs(+0x1a55db) [0x6391290d45db]
emacs(+0x5becd) [0x639128f8aecd]
/usr/lib/libc.so.6(+0x276c1) [0x7759b0cab6c1]
/usr/lib/libc.so.6(__libc_start_main+0x89) [0x7759b0cab7f9]
...
fish: Job 1, 'emacs' terminated by signal SIGSEGV (Address boundary error)
```
Not sure if this is due to this branch or is generally a bug on main (don't have time to debug or check right now).

krisbalintona:
@dakra First impressions:

It would be nice to have a mode line indicator for which input mode you're in, like EAT does. (EAT appends an indicator to the mode line lighter.)

I appreciate the echo area info on how to leave the input mode once you enter it; very useful for newcomers + those without good muscle memory for the package (yet).

Why would one enter copy mode vs emacs mode? Is copy mode meant to be a specialized emacs mode? If so, what benefit does it have over regular emacs mode? (In EAT, I would enter their emacs mode if I wanted to copy anything.)

It isn't immediately obvious what the benefit of line mode is: when should I enter it? (This one is more because I'm ignorant, not because it shouldn't exist.)

Bug in line mode?: Say I am in char mode and type some command without sending it, `ls -la`. Then I enter line mode and type anything to "complete" the command (or nothing, if I'm happy with it already), then press `RET`. The command will be sent but visually the command is doubled. That is, after my shell prompt, I'll see `ls -lals -la` if I went into line mode and just pressed `RET`. As a user, I would like whatever is shown on that shell prompt line to just be the prompt + the final command I sent in line mode. Not sure if this is possible.

dakra:
> It would be nice to have a mode line indicator for which input mode you're in, like EAT does. (EAT appends an indicator to the mode line lighter.)

There should be a mode line indicator. (for semi-char I left it blank as a default)

> Why would one enter copy mode vs emacs mode? Is copy mode meant to be a specialized emacs mode? If so, what benefit does it have over regular emacs mode? (In EAT, I would enter their emacs mode if I wanted to copy anything.)

copy mode is basically emacs mode, but copy-mode freezes the terminal while emacs-mode is live.. so e.g. if you `tail` something, copy-mode doesn't move, while in emacs-mode it continues to scroll.

> It isn't immediately obvious what the benefit of line mode is: when should I enter it? (This one is more because I'm ignorant, not because it shouldn't exist.)

I'm not an eat user, so I'm good with only semi-char+copy-mode, but many asked for eat line-mode.
There you have your normal Emacs navigation/editing bindings and don't have to use what the shell / readline provides you.

> Bug in line mode

~~Thanks, I'll have a look~~
EDIT: Should be fixed with the latest commit.

dakra:
> FYI I'm trying out this branch and calling ghostel-download-module causes my Emacs to crash

Thanks for reporting. This is fixed on latest main (by just not letting the user re-load a native module which is not supported by Emacs)

PR #79, opened 2026-04-11
build: pass -Dcpu=baseline for native x86_64 builds

ArthurHeymans:
Without this, Zig targets the build machine's native CPU. The CI
runners are modern enough to have AVX2, so the released x86_64
binary ends up with AVX2 instructions baked into compiler_rt
(rem_pio2_large, __divei4, sqrtf) via auto-vectorization. Those
are unconditional code paths with no runtime dispatch guard, so
they SIGILL immediately on Sandy Bridge and other pre-Haswell CPUs
that have AVX but not AVX2.

The Highway and simdutf SIMD paths are unaffected — they use
per-function target attributes and runtime CPUID dispatch, so they
still select AVX2 on capable CPUs. This matches what ghostty's own
nix package already does.

Tested with an AVX only Intel Sandybridge CPU.

Co-authored-by: Claude Sonnet 4.6 <claude@anthropic.com>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -41,7 +41,7 @@ if [ -n "$ZIG_TARGET" ]; then
      zig build -Demit-lib-vt=true -Doptimize=ReleaseFast $ZIG_TARGET_FLAG)
     SEARCH_DIRS="$GHOSTTY_CACHE"
 else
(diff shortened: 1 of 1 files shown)


dakra:
Thanks. I also made a new release so the new artifacts contain this PR.

Issue #76, opened 2026-04-11
handle optional dependency `evil` better


jian-lin:
Currently, installing this package from MELPA triggers an error:

1. cd $(mktemp -d)
2. HOME=$PWD emacs -Q
3. eval this
```elisp
(progn
  (require 'package)
  (add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
  (package-install 'ghostel))
```
4. see an error in the `*Compile-Log*` buffer
```
Compiling file /tmp/tmp.Hqb7js9HnK/.emacs.d/elpa/ghostel-20260410.634/ghostel-evil.el at Sat Apr 11 11:24:21 2026
ghostel-evil.el:26:11: Error: Cannot open load file: No such file or directory, evil
```

One fix [mentioned][1] by MELPA is to create a separate MELPA package for `ghostel-evil.el`.

[1]: https://github.com/melpa/melpa/blob/f092656c1fc300c4d69c762a1e6c72dd3eb657d9/README.md?plain=1#L244

dakra:
The evil package is very new. I plan to submit it to melpa.
I'm just still undecided if I want to also submit also the build files to the melpa package or not,
that's why I didn't so far so I don't have to make yet another PR later.
See https://github.com/dakra/ghostel/issues/42#issuecomment-4201839116

dakra:
I checked melpa and emacsmirror and see most evil packages start with `evil-` so I'm thinking of renameing `ghostel-evil` to `evil-ghostel`. Obviously have to do that before submitting. What do you thing @jian-lin ?

jian-lin:
> I'm thinking of renameing ghostel-evil to evil-ghostel

LGTM

dakra:
Just as a heads-up, I renamed `ghostel-evil` to `evil-ghostel`.
I added a shim but will remove it very soon, so update your config if you used `ghostel-evil` until now.

dakra:
I created a PR to melpa which adds the `evil-ghostel` package. See https://github.com/melpa/melpa/pull/9949

dakra:
jFYI, I'll soon land https://github.com/dakra/ghostel/pull/171 which refactors the directory structure to make more clean and resolve the main poin point of `evil-ghostel.el` in the same dir as the other ghostel elisp files and also fixes the missing terminfo files on melpa.
If you installed via `(use-package :vc ...` you probaby have to update your expression.
something like:

```
(use-package ghostel
[...]

PR #80, opened 2026-04-11
fix ghostel-evil.el byte-compilation without evil in load path

ArthurHeymans:
Replace the top-level (require 'evil) with declare-function/defvar
declarations and wrap the evil-define-key* keybinding setup in
(with-eval-after-load 'evil ...).

This matches the idiomatic pattern used by lsp-mode and which-key for
optional evil integration: no top-level require, declare-function with
the ext: convention to silence the byte-compiler, defvar for dynamic
variables, and with-eval-after-load to defer any top-level calls that
invoke evil at load time.

Co-authored-by: Claude Sonnet 4.6 <claude@anthropic.com>

### DIFF
--- a/ghostel-evil.el
+++ b/ghostel-evil.el
@@ -24,8 +24,22 @@
 
 ;;; Code:
 
-(require 'evil)
 (require 'ghostel)
+
+;; Silence the byte-compiler.  evil is a required runtime dependency but is
+;; not needed at compile time; users must have evil loaded before activating
+;; `ghostel-evil-mode'.
+(declare-function evil-define-key*       "ext:evil-core")
+(declare-function evil-get-register      "ext:evil-common")
+(declare-function evil-insert            "ext:evil-commands")
+(declare-function evil-refresh-cursor    "ext:evil-core")
+(declare-function evil-set-initial-state "ext:evil-core")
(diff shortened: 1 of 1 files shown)


ArthurHeymans:
This is required with nix building this package.

dakra:
Thanks.

Did you see #76.
I plan to rename it to evil-ghostel and publish it as its own package. It'll still live inside this repo though.

What do you think and would that make this PR unnecessary?

ArthurHeymans:
> Thanks.
> 
> Did you see #76.
> I plan to rename it to evil-ghostel and publish it as its own package. It'll still live inside this repo though.
> 
> What do you think and would that make this PR unnecessary?

My LLM suggested this package split too so I guess this isn't a bad idea.

dakra:
Just as a heads-up, I renamed `ghostel-evil` to `evil-ghostel`.
I added a shim but will remove it very soon, so update your config if you used `ghostel-evil` until now.

I'll open a melpa PR for it next

dakra:
Closing this as it's hopefully resolved by publishing `evil-ghostel` as it's own package.
(See melpa PR https://github.com/melpa/melpa/pull/9949 )

I don't use nix, so please re-open or create an issue if there are other problems.

PR #81, opened 2026-04-11
Make TRAMP method for OSC 7 directory tracking configurable

ArthurHeymans:
Add `ghostel-tramp-default-method' to control the TRAMP method used
when constructing remote paths from OSC 7 directory reports.  Defaults
to nil, which falls back to `tramp-default-method' instead of the
previously hardcoded "ssh".

Co-authored-by: Claude Opus 4.6 <claude@anthropic.com>


ArthurHeymans:
Should this be advertised in the readme? This works rather well with something I maintain: https://github.com/ArthurHeymans/emacs-tramp-rpc ^^

dakra:
Thanks a lot.

Yes, I would say every defcustom should also be mentioned in the readme.

dakra:
Thank you very much 🙏

PR #83, opened 2026-04-11
fix stale horizontal scroll after window resize

ArthurHeymans:
During a live frame resize, Emacs redisplays at intermediate window
sizes before the redraw timer fires.  With the old (wide) buffer
content still present, auto-hscroll-mode scrolls the window right to
keep the cursor visible.  When the timer fires and rebuilds the buffer
at the new terminal width, the stale hscroll persists because
auto-hscroll only adjusts when the cursor is outside the visible
area -- if the new cursor position happens to be visible at the old
offset, it never resets.

Fix by explicitly resetting window-hscroll to 0 after every redraw.
Terminal content is always sized to match the window, so horizontal
scrolling is never needed.

Also clamp the resize dimensions to a minimum of 1 to prevent
ghostty_terminal_resize from failing when window-max-chars-per-line
or window-body-height returns 0 for very small windows.

Co-authored-by: Claude Opus 4.6 <claude@anthropic.com>

### DIFF
--- a/ghostel.el
+++ b/ghostel.el
@@ -2243,7 +2243,7 @@ PROCESS is the shell process, WINDOWS is the list of windows."
     (when (and size (buffer-live-p buffer))
       (with-current-buffer buffer
         (when ghostel--term
-          (ghostel--set-size ghostel--term height width)
+          (ghostel--set-size ghostel--term (max 1 height) (max 1 width))
           (setq ghostel--term-rows height)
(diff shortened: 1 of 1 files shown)


dakra:
Thanks.

Could we just `(setq-local auto-hscroll-mode nil)` instead?

dakra:
Thanks again ❤️

PR #90, opened 2026-04-12
Fix cursor position for chars where char-width disagrees with terminal

### LINKED ISSUE #86: cursor position in opencode is wrong
After entering a character in opencode, the cursor should be positioned after the character, but currently it is positioned on the character.

in ghostel:
Image

in alacritty:
Image

dakra:
## Summary

- Fix cursor landing on the wrong character on lines containing box-drawing glyphs (e.g. `┃`) or other characters where Emacs' `char-width` disagrees with the terminal's column width
- Replace `move-to-column` (which relies on `char-width`) with a new `positionCursorByCell` helper that iterates the cursor row's terminal cells to compute the exact Emacs character offset, then uses `goto-char`
- The new approach is also faster since it avoids Elisp display-width scanning (~40% faster in benchmarks with non-ASCII content)

Fixes #86

## Test plan

- [x] `zig build check` compiles cleanly
- [x] All 56 Elisp tests pass
- [x] Benchmarked before/after: no regression, actually faster across all scenarios
- [ ] Needs verification on Linux pgtk with CJK locale (the reporter's environment)

### DIFF
--- a/src/render.zig
+++ b/src/render.zig
@@ -750,6 +750,70 @@ fn insertScrollbackRange(
 }
 
 
+/// Convert a terminal column to an Emacs character offset by iterating
+/// the row's cells.  Returns `true` and positions point on success;
+/// `false` if the cell data is unavailable (caller should fall back to
+/// `move-to-column`).
+///
+/// This avoids relying on Emacs' `char-width`, which can disagree with
+/// the terminal's column width for certain characters (e.g. box-drawing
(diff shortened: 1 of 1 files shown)


jixiuf:
I tried this,  and it is fixed by this PR.

dakra:
Great. Thanks for testing and the wide-character hint :)

PR #93, opened 2026-04-13
Build with Zig and vendor emacs-module.h                                                                                                               

kiennq:
- replace build.sh with zig build across local build entrypoints, Emacs compile commands, CI, release automation, and bench helpers                   
   - vendor include/emacs-module.h so the default build no longer depends on locally installed Emacs headers                                             
   - keep header override support via EMACS_INCLUDE_DIR and EMACS_BIN_DIR                                                                                
   - remove the wrapper-header path and use direct emacs-module.h inclusion in the Zig build                                                             
   - update docs and tests for the new build flow


dakra:
Thank you very much 🙏

PR #94, opened 2026-04-13
Decouple module downloads from package version

kiennq:
Allow native module downloads to target the minimum supported release or an explicit release tag without depending on the package version.

Keep the GitHub release URL customizable for forked release hosting.


dakra:
Thanks again 🙏

PR #108, opened 2026-04-15
Add Windows ConPTY support

kiennq:
- ConPTY pseudoconsole backend for Windows terminal I/O
- Transport abstractions (ghostel--process-send/live-p/set-window-size) that dispatch to ConPTY or PTY depending on the active backend
- Windows-specific process startup skipping the shell wrapper
- ConPTY module build via emacs-util-mods dependency
- Release artifacts packaged as per-platform .tar.xz archives containing ghostel-module + conpty-module (Windows)
- Download/extract flow for .tar.xz release archives
- Windows entries in CI build and test matrix
- 7 new ConPTY tests + extract-module-archive test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,13 @@ jobs:
           version: 0.15.2
 
       - name: Run pure-Zig unit tests
-        run: zig build test
+        shell: bash
+        run: |
+          target_args=()
+          if [ "${{ runner.os }}" = "Windows" ]; then
+            target_args=(-Dtarget=x86_64-windows-gnu)
+          fi
+          zig build test "${target_args[@]}"
 
   test:
     runs-on: ubuntu-latest
@@ -287,6 +293,7 @@ jobs:
   test-native-windows:
     needs: build-native
     runs-on: windows-latest
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
 
@@ -310,7 +317,8 @@ jobs:
         run: |
           touch ghostel-module.dll
           git clone --depth 1 https://github.com/emacs-compat/compat.git /tmp/compat
-          make -j$(getconf _NPROCESSORS_ONLN) \
+          make \
+            EMACS="timeout 60s emacs" \
             EMACSFLAGS="-L /tmp/compat" \
             test-native
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,6 +164,10 @@ jobs:
             platform: x86_64-windows
             suffix: .dll
             target: x86_64-windows-gnu
+          - os: windows-latest
+            platform: aarch64-windows
+            suffix: .dll
+            target: aarch64-windows-gnu
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,10 @@ jobs:
             platform: x86_64-freebsd
             suffix: .so
             target: x86_64-freebsd
+          - os: windows-latest
+            platform: x86_64-windows
+            suffix: .dll
+            target: x86_64-windows-gnu
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
@@ -178,6 +182,7 @@ jobs:
           cache-key: ${{ matrix.platform }}-zig-0.15.2
 
       - name: Build native module
+        shell: bash
         run: |
           if [ -n "${{ matrix.target }}" ]; then
             zig build --prefix . -Dtarget=${{ matrix.target }} -Doptimize=ReleaseFast -Dcpu=baseline
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,13 @@ jobs:
         run: make package-lint checkdoc docquotes
 
   test-zig:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
 
@@ -185,13 +191,24 @@ jobs:
           version: 0.15.2
           cache-key: ${{ matrix.platform }}-zig-0.15.2
 
+      - name: Download ConPTY runtime
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          Invoke-WebRequest -Uri https://www.nuget.org/api/v2/package/Microsoft.Windows.Console.ConPTY/1.24.260512001 -OutFile conpty.zip
+          Expand-Archive -LiteralPath conpty.zip -DestinationPath conpty-package -Force
+
       - name: Build native module
         shell: bash
         run: |
+          conpty_args=()
+          if [ "${{ runner.os }}" = "Windows" ]; then
+            conpty_args=(-Dwindows-conpty-package-dir=conpty-package)
+          fi
           if [ -n "${{ matrix.target }}" ]; then
-            zig build --prefix . -Dtarget=${{ matrix.target }} -Doptimize=ReleaseFast -Dcpu=baseline
+            zig build --prefix . -Dtarget=${{ matrix.target }} -Doptimize=ReleaseFast -Dcpu=baseline "${conpty_args[@]}"
           else
-            zig build --prefix . -Doptimize=ReleaseFast -Dcpu=baseline
+            zig build --prefix . -Doptimize=ReleaseFast -Dcpu=baseline "${conpty_args[@]}"
           fi
 
       - name: Upload module artifact
@@ -201,6 +218,9 @@ jobs:
           path: |
             ghostel-module${{ matrix.suffix }}
             ghostel-module.version
+            conpty.dll
+            x64/OpenConsole.exe
+            arm64/OpenConsole.exe
 
   test-native:
     needs: build-native
@@ -227,6 +247,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
       - uses: cachix/install-nix-action@v31
         with:
           github_access_token: ${{ github.token }}
@@ -257,6 +280,36 @@ jobs:
 
       - name: Run native module tests
         run: |
+          make -j$(getconf _NPROCESSORS_ONLN) \
+            EMACSFLAGS="-L /tmp/compat" \
+            test-native
+
+  test-native-windows:
+    needs: build-native
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: UCRT64
+          update: true
+          install: >-
+            git
+            make
+            mingw-w64-ucrt-x86_64-emacs
+            mingw-w64-ucrt-x86_64-python
+
+      - name: Download native module
+        uses: actions/download-artifact@v4
+        with:
+          name: ghostel-module-x86_64-windows
+
+      - name: Run Windows native module tests
+        shell: msys2 {0}
+        run: |
+          touch ghostel-module.dll
+          git clone --depth 1 https://github.com/emacs-compat/compat.git /tmp/compat
           make -j$(getconf _NPROCESSORS_ONLN) \
             EMACSFLAGS="-L /tmp/compat" \
             test-native
--- a/build.zig
+++ b/build.zig
@@ -71,6 +71,12 @@ pub fn build(b: *std.Build) void {
     const copy_version_step = b.addInstallFile(version_file, "ghostel-module.version");
     b.getInstallStep().dependOn(&copy_version_step.step);
 
+    if (target_os == .windows) {
+        if (b.option([]const u8, "windows-conpty-package-dir", "Unpacked Microsoft.Windows.Console.ConPTY NuGet package directory")) |dir| {
+            installWindowsConptyRuntime(b, target.result.cpu.arch, dir);
+        }
+    }
+
     // ----------------------------------------------------------------
     // `zig build test` — pure-Zig unit tests.
     //
@@ -161,4 +167,41 @@ fn moduleOutputName(target_os: std.Target.Os.Tag) []const u8 {
         .windows => "ghostel-module.dll",
         else => "ghostel-module.so",
     };
+}
+
+fn installWindowsConptyRuntime(b: *std.Build, arch: std.Target.Cpu.Arch, package_dir: []const u8) void {
+    const runtime_arch = windowsConptyRuntimeArch(arch) orelse return;
+    const conpty_path = b.pathJoin(&.{ package_dir, "runtimes", b.fmt("win-{s}", .{runtime_arch}), "native", "conpty.dll" });
+    const copy_conpty = b.addInstallFile(.{ .cwd_relative = conpty_path }, "conpty.dll");
+    b.getInstallStep().dependOn(&copy_conpty.step);
+
+    switch (arch) {
+        .x86 => {
+            installWindowsOpenConsole(b, package_dir, "x86");
+            installWindowsOpenConsole(b, package_dir, "x64");
+            installWindowsOpenConsole(b, package_dir, "arm64");
+        },
+        .x86_64 => {
+            installWindowsOpenConsole(b, package_dir, "x64");
+            installWindowsOpenConsole(b, package_dir, "arm64");
+        },
+        .aarch64 => installWindowsOpenConsole(b, package_dir, "arm64"),
+        else => {},
+    }
+}
+
+fn windowsConptyRuntimeArch(arch: std.Target.Cpu.Arch) ?[]const u8 {
+    return switch (arch) {
+        .x86 => "x86",
+        .x86_64 => "x64",
+        .aarch64 => "arm64",
+        else => null,
+    };
+}
+
+fn installWindowsOpenConsole(b: *std.Build, package_dir: []const u8, host_arch: []const u8) void {
+    const source = b.pathJoin(&.{ package_dir, "build", "native", "runtimes", host_arch, "OpenConsole.exe" });
+    const dest = b.fmt("{s}/OpenConsole.exe", .{host_arch});
+    const copy = b.addInstallFile(.{ .cwd_relative = source }, dest);
+    b.getInstallStep().dependOn(&copy.step);
 }
--- a/build.zig
+++ b/build.zig
@@ -53,6 +53,9 @@ pub fn build(b: *std.Build) void {
             lib.setVersionScript(b.path("symbols.map"));
         }
     }
+    if (target_os == .windows) {
+        lib.linkSystemLibrary("kernel32");
+    }
 
     const copy_step = b.addInstallFile(
         lib.getEmittedBin(),
@@ -155,6 +158,7 @@ fn dirHasEmacsModuleHeader(allocator: std.mem.Allocator, dir: []const u8) bool {
 fn moduleOutputName(target_os: std.Target.Os.Tag) []const u8 {
     return switch (target_os) {
         .macos => "ghostel-module.dylib",
+        .windows => "ghostel-module.dll",
         else => "ghostel-module.so",
     };
 }
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -4378,6 +4378,7 @@ writes a final exit status before closing it."
                 :noquery t))
          (pid (ghostel--spawn-native-process ghostel--term command pipe)))
     (setq ghostel--pid pid)
+    (process-put pipe 'ghostel--native-pid pid)
 
     (set-process-sentinel pipe
                           (lambda (process event)
@@ -4385,10 +4386,21 @@ writes a final exit status before closing it."
                             ;; If Emacs deletes it before normal exit, make sure
                             ;; the child is not left running.  After normal exit
                             ;; the reaper has already waited, so this is a no-op.
-                            (signal-process pid 9)
+                            (signal-process
+                             (process-get process 'ghostel--native-pid) 9)
                             (ghostel--sentinel process event)))
     (add-hook 'kill-buffer-hook #'ghostel--kill-native-process-hook nil t)
     pipe))
+
+(defun ghostel--kill-native-processes-on-exit ()
+  "Force native children to exit before Emacs disables process sentinels."
+  (dolist (process (process-list))
+    (let ((pid (and (process-live-p process)
+                    (process-get process 'ghostel--native-pid))))
+      (when pid
+        (ignore-errors (signal-process pid 9))))))
+
+(add-hook 'kill-emacs-hook #'ghostel--kill-native-processes-on-exit)
 
 (defun ghostel--kill-native-process-hook ()
   "Detach the native event pipe and request child termination.
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -228,6 +228,7 @@ README for the full design and per-call escape hatch."
 
 (defcustom ghostel-tramp-shells
   '(("ssh" login-shell)
+    ("sshx" login-shell)
     ("scp" login-shell)
     ("docker" "/bin/sh"))
   "Shell to use for remote TRAMP connections, per method.
@@ -3120,22 +3121,19 @@ default."
   "Return non-nil if the foreground program looks like it's reading a password.
 Two arms:
 
-  - libghostty heuristic (`ghostel--pty-password-input-p'): the current
+  - Zig heuristic (`ghostel--pty-password-input-p'): the current
     PTY is in canonical mode with echo off.  Catches local sudo, ssh's
     own password prompt, gpg, etc.
 
   - cursor-row regex (`ghostel-password-prompt-regex', defaulting to
-    `comint-password-prompt-regexp').  Used only when the libghostty
-    heuristic returns nil AND `ghostel--remote-shell-p' indicates a
-    remote shell - the case where the local pty is in raw mode for ssh
-    forwarding and the remote pty's canonical+!echo isn't visible locally.
+    `comint-password-prompt-regexp').  Used when `ghostel--remote-shell-p'
+    indicates a remote shell, where the remote pty's canonical+!echo state
+    isn't visible through a local pty probe.
 
 Returns nil on miss, or a symbol naming the arm on hit (`zig' or`regex')."
-  (cond
-   ((ghostel--pty-password-input-p ghostel--term) 'zig)
-   ((and (ghostel--remote-shell-p)
-         (ghostel--password-regex-matches-cursor-row-p))
-    'regex)))
+  (if (ghostel--remote-shell-p)
+      (when (ghostel--password-regex-matches-cursor-row-p) 'regex)
+    (when (ghostel--pty-password-input-p ghostel--term) 'zig)))
 
 (defun ghostel--password-regex-matches-cursor-row-p ()
   "Return non-nil if the cursor row looks like a password prompt.
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -4311,13 +4311,15 @@ Local PROGRAM is resolved to an absolute path before backend dispatch.
 Local buffers use the native PTY path when `ghostel-use-native-pty'
 is non-nil; remote (REMOTE-P) buffers always go through Emacs so
 TRAMP can manage the remote shell."
-  (let ((program (if remote-p
-                     program
-                   (ghostel--resolve-local-executable program))))
-    (setq ghostel--process
-          (if (and ghostel-use-native-pty (not remote-p))
-              (ghostel--spawn-via-native (cons program program-args))
-            (ghostel--spawn-via-emacs program program-args remote-p)))))
+  (let* ((program (if remote-p
+                      program
+                    (ghostel--resolve-local-executable program)))
+         (process (if (and ghostel-use-native-pty (not remote-p))
+                      (ghostel--spawn-via-native (cons program program-args))
+                    (ghostel--spawn-via-emacs program program-args remote-p))))
+    (when (processp process)
+      (process-put process 'adjust-window-size-function #'ignore))
+    (setq ghostel--process process)))
 
 (defun ghostel--spawn-via-emacs (program program-args &optional remote-p)
   "Spawn PROGRAM with PROGRAM-ARGS through Emacs process machinery.
--- a/src/ConPtyProcess.zig
+++ b/src/ConPtyProcess.zig
@@ -1,6 +1,8 @@
 const std = @import("std");
 const Allocator = std.mem.Allocator;
 
+const backend_types = @import("backend_types.zig");
+
 const c = @cImport({
     @cInclude("windows.h");
 });
@@ -13,23 +15,27 @@ const ResizePseudoConsoleFn = *const fn (HPCON, c.COORD) callconv(.winapi) c.HRE
 const ClosePseudoConsoleFn = *const fn (HPCON) callconv(.winapi) void;
 
 const PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE: usize = 0x00020016;
+const PIPE_REJECT_REMOTE_CLIENTS: c.DWORD = 0x00000008;
+const GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT: c.DWORD = 0x00000002;
+const GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS: c.DWORD = 0x00000004;
 const READ_BUFFER_SIZE = 64 * 1024;
+const WRITE_CHUNK_SIZE = 4096;
 
+const OverlappedPipeEnd = enum { read, write };
+
+alloc: Allocator,
 hpc: HPCON = null,
 pty_input: c.HANDLE = c.INVALID_HANDLE_VALUE,
 pty_output: c.HANDLE = c.INVALID_HANDLE_VALUE,
-command_event: c.HANDLE = c.INVALID_HANDLE_VALUE,
+interrupt_event: c.HANDLE = c.INVALID_HANDLE_VALUE,
 output_read_event: c.HANDLE = c.INVALID_HANDLE_VALUE,
+input_write_event: c.HANDLE = c.INVALID_HANDLE_VALUE,
 shell_process: c.HANDLE = c.INVALID_HANDLE_VALUE,
-running: std.atomic.Value(bool) = .init(true),
 pid: i64 = -1,
 
 var create_pseudo_console: ?CreatePseudoConsoleFn = null;
 var resize_pseudo_console: ?ResizePseudoConsoleFn = null;
 var close_pseudo_console: ?ClosePseudoConsoleFn = null;
-var pipe_name_counter = std.atomic.Value(u32).init(1);
-
-pub const ProcessParams = @import("ProcessParams.zig");
 
 pub const EventWriter = struct {
     pub const Fd = c_int;
@@ -85,6 +91,8 @@ pub const EventWriter = struct {
     fn notifyCrtProviderForImport(dll_name: []const u8, symbol_name: []const u8) ?NotifyCrtProvider {
         if (!std.mem.eql(u8, symbol_name, "_dup")) return null;
 
+        // Emacs' `open_channel' returns a CRT fd owned by Emacs, so event
+        // writes must go through the same mainstream CRT family Emacs imports.
         if (std.ascii.eqlIgnoreCase(dll_name, "msvcrt.dll")) return .msvcrt;
         if (std.ascii.eqlIgnoreCase(dll_name, "ucrtbase.dll")) return .ucrt;
         // Newer Emacs builds may link the UCRT through api-ms-win-crt
@@ -172,17 +180,23 @@ pub const EventWriter = struct {
     }
 };
 
-pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: ProcessParams) !Self {
+pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: backend_types.ProcessParams) !Self {
     try initApi();
 
-    var self: Self = .{};
-    self.command_event = c.CreateEventW(null, c.TRUE, c.FALSE, null);
-    if (self.command_event == null) return error.CreateEventFailed;
-    errdefer self.closeConPtyHandles();
+    var self: Self = .{ .alloc = alloc };
+
+    self.interrupt_event = c.CreateEventW(null, c.TRUE, c.FALSE, null);
+    if (self.interrupt_event == null) return error.CreateEventFailed;
+    errdefer closeConPtyHandles(&self);
+
     self.output_read_event = c.CreateEventW(null, c.TRUE, c.FALSE, null);
     if (self.output_read_event == null) return error.CreateEventFailed;
-    try self.createConPty(initial_rows, initial_cols);
-    try self.spawnChild(alloc, params);
+
+    self.input_write_event = c.CreateEventW(null, c.TRUE, c.FALSE, null);
+    if (self.input_write_event == null) return error.CreateEventFailed;
+
+    try createConPty(&self, initial_rows, initial_cols);
+    try spawnChild(&self, params);
 
     return self;
 }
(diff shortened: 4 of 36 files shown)


dakra:
Thank you very much for this. Imho looks already really good.
I let claude review it and also added that I would like a good separation between the existing and the windows code so I can revert it if there should be something major wrong/buggy with the windows code and I'm not able to maintain it (hopefully not ;) )

So here is the claude review after chatting a bit.
Don't take it too serious. Make changes as you see it makes sense.
I'm not too worried about the separation as looking at it now is already pretty clean imho,
but having it separate and in one place as much as possible is preferred.
Just use your judgement ;)

Here the review:
------

# Review

Thanks for this — the ConPTY work is well-structured overall and the transport abstraction is the right shape. Before merging I'd like a few changes to make the Windows path more isolated (so it's easy to revert as a unit if Windows-specific bugs surface), plus a few smaller fixes from the review. Detailed list below; an LLM should be able to crank through these mechanically.

## Required: Windows isolation

The goal is that reverting Windows support is "delete one section + one dispatch line + one bootstrap block", with no diff churn in shared Unix code paths.

### 1. Cluster all Windows-only elisp under one section

Group the following under a single `;;; Windows ConPTY backend` section header in `ghostel.el` (currently they're scattered):

- `conpty--*` `declare-function` block
- `ghostel--conpty-notify-pipe` defvar
- `ghostel--conpty-module-file-path`
- `ghostel--ensure-conpty-loaded`
- `ghostel--native-runtime-ready-p`
- `ghostel--conpty-active-p`, `ghostel--process-live-p`, `ghostel--process-send`, `ghostel--process-set-window-size`, `ghostel--conpty-filter`
- `ghostel--start-process-windows` (see #2)
- `ghostel--extract-module-archive`, `ghostel--publish-downloaded-module-archive`, `ghostel--replace-module-file` if they're Windows-only — if they're shared with the new tar.xz flow, leave them in the download section.

### 2. Factor Windows process startup out of `ghostel--start-process`

Right now `ghostel--start-process` is a fork inside one big `let*`: env got split into `env-overrides` + `process-environment`, the `shell-command` form has an `if windows-nt`, and `set-process-coding-system` / `set-process-window-size` / query-flag are wrapped in `unless windows-nt`. The Unix body no longer matches its pre-PR shape, so a revert isn't a clean delete.

Please restructure so:

- `ghostel--start-process-windows` (new defun, in the Windows section) contains all Windows-specific setup: `make-pipe-process`, `process-put`, `conpty--init`, query-flag, the Windows-specific `shell-command` (just `(list shell)`), and the env construction it needs.
- `ghostel--start-process` becomes:
  ```elisp
  (if (eq system-type 'windows-nt)
      (ghostel--start-process-windows ...)
    <original Unix body, byte-identical to pre-PR>)
  ```
- The Unix body should not have `unless windows-nt` wrappers around `set-process-coding-system` / `set-process-window-size` / `set-process-query-on-exit-flag` / `process-put` — those guards go away because that whole branch only runs on Unix.
- Drop the redundant inner `(setq ghostel--process proc)` and inner `set-process-query-on-exit-flag` in the Windows branch — `:noquery t` already covers the latter and the outer setq covers the former.

### 3. Keep transport helpers as permanent thin wrappers

`ghostel--process-send`, `ghostel--process-live-p`, `ghostel--process-set-window-size` should stay even if ConPTY is later reverted — that way the ~10 call-site rewrites in `ghostel.el` (`process-send-string` → `ghostel--process-send`, `process-live-p` → `ghostel--process-live-p`) don't have to be undone. On non-Windows, the helpers are one-line forwarders. Please add a brief comment on each helper noting they're permanent wrappers around the underlying primitive, not a Windows-only abstraction.

### 4. Bootstrap: avoid nesting the ConPTY load inside the existing `condition-case`

Currently the new `(when (eq system-type 'windows-nt) (condition-case err (ghostel--ensure-conpty-loaded dir) ...))` is nested inside the existing `condition-case err` for `module-load`, which shadows `err`. Please move it to a separate top-level `(when (eq system-type 'windows-nt) ...)` block immediately after the existing bootstrap so it's a clean append.

In `ghostel-toggle`, the swap from `(fboundp 'ghostel--new)` to `(ghostel--native-runtime-ready-p)` is fine — keep it. `ghostel--native-runtime-ready-p` already short-circuits the conpty check to `t` on non-Windows.

## Required: Restore macOS strip step

The release workflow rewrite dropped `strip -x` on macOS without replacement. `strip -x` removes non-exported symbols from the dylib and meaningfully shrinks the artifact. It's orthogonal to the tar.xz packaging change.

Please restore it as a pre-tar step in `.github/workflows/release.yml`:

```yaml
- name: Strip non-exported symbols (macOS)
  if: runner.os == 'macOS'
  run: strip -x zig-out/bin/ghostel-module${{ matrix.suffix }}
```

(Place it before the "Package release artifact" step.)

The universal `.tar.xz` packaging is good — keep it for all platforms; xz compresses `.so`/`.dylib` well and saves bandwidth. The conpty bundling logic for Windows-only is correct.

## Smaller fixes

### 5. Resize ordering in the ConPTY branch

In `ghostel--window-adjust-process-window-size`, the new `(ghostel--process-set-window-size process height width)` runs *after* `(ghostel--delayed-redraw buffer)`, so the buffer is redrawn against the new term size before the ConPTY child has been told to resize. Please call `ghostel--process-set-window-size` **before** the redraw — or, ideally, before `ghostel--set-size` — so the child knows the new size first.

### 6. Simplify `ghostel--conpty-module-file-path`

The current construction wraps `expand-file-name` around `file-name-as-directory` around another `expand-file-name`. Simpler:

```elisp
(defun ghostel--conpty-module-file-path (&optional dir)
  "Return the stable conpty-module path in DIR."
  (expand-file-name (concat "conpty-module" module-file-suffix)
                    (or dir (file-name-directory (or load-file-name
                                                     (locate-library "ghostel")
                                                     buffer-file-name)))))
```

### 7. `addWindowsRuntimeLibraries` — clarify or drop the MSVC branch

The SDK lookup only runs when `abi == .msvc`, but the entire CI matrix builds with `*-windows-gnu`. Either:

- Drop the `if (rt.abi != .msvc) return;` block and everything below it (simpler, matches what's actually exercised), or
- Leave it but add a comment explaining it's there for future MSVC builds.

### 8. Live ConPTY test gated off CI

`ghostel-test-conpty-init-keeps-shell-alive-on-windows` has `(skip-unless (not (getenv "GITHUB_ACTIONS")))`, so the only end-to-end ConPTY assertion never runs in CI — the Windows job currently just smoke-tests that the module loads. Please investigate the underlying reason (cmd.exe startup timing? sleep/poll race?) and either fix it or document in a comment why it has to be skipped, with a TODO.

### 9. Tar extraction note

`tar xJf … -C dest` doesn't block path traversal in archive entries. The source is your own GitHub release so this is acceptable, but please add a one-line comment in `ghostel--extract-module-archive` noting the trust assumption, so future maintainers don't repurpose it for untrusted archives.

## Nice-to-have (optional)

- Consider `strip --strip-unneeded` for Linux release artifacts symmetrically with macOS — wasn't in the original workflow either, so up to you.
- The `ghostel-test-extract-module-archive-uses-tar-xf` test is great; consider adding a complementary test that verifies `ghostel--publish-downloaded-module-archive` only copies `conpty-module.dll` on Windows, not on Linux/macOS.

## Summary of the revert story after these changes

If Windows ever needs to be ripped out:

1. Delete the `;;; Windows ConPTY backend` section in `ghostel.el`.
2. Replace the `(if (eq system-type 'windows-nt) (ghostel--start-process-windows ...) <unix>)` dispatch with just `<unix>`.
3. Delete the `(when (eq system-type 'windows-nt) ...)` bootstrap block.
4. Delete the `if (target_os == .windows)` blocks in `build.zig` and the `emacs_util_mods` dep in `build.zig.zon`.
5. Remove Windows rows from CI matrices.

Transport helpers stay; tar.xz packaging stays; no shared code paths get touched. That's the goal.

kiennq:
Thanks, the review is helpful. I will try to address most of them. However for the strip, the Zig build system already handles stripping — build.zig sets `.strip = true` for release builds, plus enables link_gc_sections, link_function_sections, link_data_sections, and dead_strip_dylibs. This produces the same result as strip -x (removes non-exported symbols) without needing an external tool step. The original strip -x in the workflow was compensating for a build that didn't strip — with the Zig flags it's redundant.

dakra:
> However for the strip, the Zig build system already handles stripping

At least for MacOS that's not completely true.
I build with `zig build -Doptimize=ReleaseFast -Dcpu=baseline`
and then did `strip -x`:

| State | Size | Defined symbols | Total symbol table |
|---|---|---|---|
| Zig build only | 1,556,536 bytes (1,520 KB) | 3,879 | 4,084 |
| After `strip -x` | 1,381,000 bytes (1,349 KB) | 389 | 594 |
| **Savings** | **175,536 bytes (171 KB, 11.3%)** | -3,490 | -3,490 |

`.strip = true` in Zig strips debug info but leaves the symbol table mostly intact, so it's not equivalent to `strip -x`.

kiennq:
@dakra, please let me know if you want me to fix anything else

dakra:
@kiennq Thank you very much. Looks already really good.

3 quick thoughts (without consulting AI ;) ):

Iirc I did the strip -x only for Mac because it was only needed there.
On Linux they where already stripped, you added it for all systems and with strip -x .. || strip ...
not sure if that change is needed (and relevant for the windows build).

There are a few unrelated indent changes

What do you think if we just copy your conpty.zig in here? It's only 560 lines and probably doesn't change very often. would have the advantage to have 1 less external dependency and also we would not need a few workarounds (e.g. tar) because we have 2 .dll files for windows.

I would have a closer look at this at the weekend.

Thanks again ❤️

kiennq:
> What do you think if we just copy your conpty.zig in here? It's only 560 lines and probably doesn't change very often. would have the advantage to have 1 less external dependency and also we would not need a few workarounds (e.g. tar) because we have 2 .dll files for windows.

The conpty is a module on its own and independent from `ghostel` so I would like to keep it separated. It can be used by emacs-vterm as well, so pretty versatile.

dakra:
> The conpty is a module on its own and independent from ghostel so I would like to keep it separated. It can be used by emacs-vterm as well, so pretty versatile.

Sure, I don't want to move the emacs-utils repo (/conpty) in here. Just as it is one small file, kinda vendor it here similar to emacs.h or I'll soon add another small PNG decoder for the kitty graphics support.
That way we would save a few extra steps that comes with having to deal with 2 .dll files on windows.
Not sure what the disadvantages would be by vendoring it in vs having it as extra dep?
(And just to be clear, I don't have too strong of an opinion, just discussing..)

kiennq:
> Sure, I don't want to move the emacs-utils repo (/conpty) in here. Just as it is one small file, kinda vendor it here similar to emacs.h or I'll soon add another small PNG decoder for the kitty graphics support.

Since the conpty will expose some module functions, would it be okay for them to clobber the `emacs_module_init` in ghostel-module?
I don't have much concern about one or multiple files on Windows since we will still treat them all the same, unpack -> move the currently in-used binaries to backup -> deploy the new one. That's the only way we can download the new module while the current ghostel is running

dakra:
After thinking and talking with Claude a bit more about it, I personally would favour to put the conpty stuff just under src/conpty or so. I couldn't see any downside and only a few improvements (one dll download, no tar calls, no ghostel--ensure-conpty-loaded (because if ghostel is loaded, so is contpy on windows), etc)

> Since the conpty will expose some module functions, would it be okay for them to clobber the emacs_module_init in ghostel-module?

sure, you can make the changes necessary.

Also sorry for always changing the codebase under you. ghostel is still in heavy development,
so hopefully we can get this merged rather sooner than later so we don't always have to rewrite the rebase on top of a changed main :)

ISouthRain:
@dakra @kiennq 

Hello, every one.

What happened?
 
Making ghostel support Windows seems to be going smoothly, where `kiennq/ghostel` is merged and updated daily from `dakra/ghostel`, and runs normally on Windows.

Not sure why this PR stalled?

lambertlulala:
Very appreciate @dakra @kiennq ,

I use https://github.com/kiennq/ghostel on Windows, and it can perfectly resolve some issues that both eat and vterm have.
It's great to have native terminal in native Windows Emacs, so I don't have to run Emacs inside WSL.

emil-e:
@kiennq Just got notified about this one. Actually had you in mind when implementing the native POSIX backend. I'd really like to be able to get this into mainline Ghostel. With the new native PTY setup it should be a lot easier to get that to work.

Are you finished with your changes here? Please tell me when you are and I can start reviewing it so maybe we can iterate and finally have Ghostel officially support Windows.

kiennq:
Thanks @emil-e, please help review the latest change.

lisp/ghostel.el
```diff
@@ -5106,13 +5121,16 @@ Local PROGRAM is resolved to an absolute path before backend dispatch.
...
-            (ghostel--spawn-via-emacs program program-args remote-p)))))
+  (let* ((program (if remote-p
+                      program
+                    (ghostel--resolve-local-executable program)))
+         (process (if (and ghostel-use-native-pty (not remote-p))
+                      (ghostel--spawn-via-native (cons program program-args))
+                    (ghostel--spawn-via-emacs program program-args remote-p))))
+    (when (processp process)
+      (process-put process 'adjust-window-size-function
+                   #'ghostel--window-adjust-process-window-size))
```
emil-e: What's the reason for this? The intention is to not use the `adjust-window-size-function` but to use the `window-size-change-functions` hook.

I also only see a declaration, the function doesn't seem to exist.

src/GhostelTerm.zig
```diff
@@ -15,6 +15,8 @@ const utils = @import("utils.zig");
 const parseHexColor = utils.parseHexColor;
 const PtyProcess = @import("PtyProcess.zig");
 const NativeProcess = @import("NativeProcess.zig");
+const ChannelFd = NativeProcess.ChannelFd;
+const ChannelPid = i64;
```
emil-e: The PID is for the process, not the channel, so maybe just `Pid` or `ProcessPid`.

src/NativeProcess.zig
```diff
@@ -141,53 +136,19 @@ fn run(self: *Self) void {
...
-            .events = posix.POLL.IN,
-            .revents = undefined,
-        },
-        .{
-            .fd = self.wake_pipe[0],
-            .events = posix.POLL.IN,
-            .revents = undefined,
-        },
-    };
+    var buf: [64 * 1024]u8 = undefined;
```
emil-e: Using a fixed bufer like this puts an upper bound on how much you can read in one go without `notifyVtUpdate` below. This will reduce performance on large multi-megabyte output chunks. A better API in `process.read` would be to rename it to `drain` and have it write directly to the stream instead. That would make the API also more flexible.

src/PtyProcess.zig
```diff
@@ -1,192 +1,46 @@
...
-        }
-    }
-
-    pub fn closeReplica(self: *@This()) void {
-        if (self.replica_fd != -1) {
-            posix.close(self.replica_fd);
-            self.replica_fd = -1;
-        }
-    }
+backend: Backend,
```
emil-e: Since this is only a wrapper around the respective backend implementation, we can just switch the implementation in `NativeProcess` instead and skip this wrapper.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+        const close_proc = c.GetProcAddress(module, "_close") orelse return error.NotifyCrtUnavailable;
+        return .{
+            .write = @ptrCast(write_proc),
+            .close = @ptrCast(close_proc),
+        };
+    }
+};
+
+pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: ProcessParams) !Self {
+    if (!(try initApi())) return error.MissingConPty;
```
emil-e: `initApi` can return `error.MissingConPty` directly instead.

src/NativeProcess.zig
```diff
@@ -117,22 +112,22 @@ fn writeEventLispString(self: *Self, str: []const u8) !void {
 }
 
 fn run(self: *Self) void {
-    self.blockSigpipe();
-    defer drainSigpipe();
+    self.event_writer.blockSigpipe();
```
emil-e: Since only POSIX needs these, how about renaming the `blockSigpipe` and `drainSigpipe` functions to `onThreadEnter` and  `onThreadExit`? Or similar.

src/PosixPtyProcess.zig
```diff
@@ -0,0 +1,314 @@
...
+    }
+}
+
+pub fn replicaName(self: *Self) []const u8 {
+    return self.pty.replicaName();
+}
+
+pub fn deinitAndWait(self: *Self) u8 {
+    self.pty.deinit();
+    if (self.wake_pipe[0] != -1) {
```
emil-e: I don't think we should have -1 checks since that could only happen if you call `deinitAndWait` twice, which should never happen.

build.zig
```diff
@@ -158,6 +173,55 @@ fn dirHasEmacsModuleHeader(allocator: std.mem.Allocator, dir: []const u8) bool {
 fn moduleOutputName(target_os: std.Target.Os.Tag) []const u8 {
     return switch (target_os) {
         .macos => "../ghostel-module.dylib",
+        .windows => "../ghostel-module.dll",
         else => "../ghostel-module.so",
     };
 }
+
+fn explicitWindowsAbiTarget(
```
emil-e: Could you explain what these functions are doing? I'm not great with Zig build files.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    }
+
+    fn pop(self: *ResizeQueue) ?ResizeRequest {
+        if (!self.pending) return null;
+        self.pending = false;
+        return self.request;
+    }
+};
+
+const State = struct {
```
emil-e: Any reason why this is a separate struct rather than just being inserted into `Self`?

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+
+            const import_by_name = ptrFromRva(c.IMAGE_IMPORT_BY_NAME, base, @intCast(address));
+            const name_addr = @intFromPtr(import_by_name) + @offsetOf(c.IMAGE_IMPORT_BY_NAME, "Name");
+            const import_name: [*:0]const u8 = @ptrFromInt(name_addr);
+            if (std.mem.eql(u8, std.mem.span(import_name), symbol_name)) return true;
+        }
+        return false;
+    }
+
+    fn findNotifyCrtProviderInImage(module: c.HMODULE) ?NotifyCrtProvider {
```
emil-e, line 135: Are there Emacs builds that use `ucrtbase`? My impression was that Emacs would only build with `msvcrt`. If both are possible, we should obviously detect both, but otherwise we could simplify.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+
+    const NotifyCrtWriteFn = *const fn (c_int, ?*const anyopaque, c_uint) callconv(.c) c_int;
+    const NotifyCrtCloseFn = *const fn (c_int) callconv(.c) c_int;
+    const NotifyCrt = struct {
+        write: NotifyCrtWriteFn,
+        close: NotifyCrtCloseFn,
+    };
+
+    fd: Fd,
+    notify_crt: NotifyCrt,
```
emil-e: This should be the same every time so I'm not sure we need to resolve this for every channel, it could be done once globally instead.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    state.* = .{ .alloc = alloc };
+
+    c.InitializeCriticalSection(&state.resize_lock);
+    var resize_lock_initialized = true;
+    errdefer if (resize_lock_initialized) c.DeleteCriticalSection(&state.resize_lock);
+
+    state.resize_event = c.CreateEventW(null, c.FALSE, c.FALSE, null) orelse return error.CreateResizeEventFailed;
+    errdefer {
+        _ = c.CloseHandle(state.resize_event);
+        state.resize_event = c.INVALID_HANDLE_VALUE;
```
emil-e: A failed init doesn't have to produce a valid state so no need to reset the resize_event here.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    hpc: HPCON = null,
+    pty_input: c.HANDLE = c.INVALID_HANDLE_VALUE,
+    pty_output: c.HANDLE = c.INVALID_HANDLE_VALUE,
+    shell_process: c.HANDLE = c.INVALID_HANDLE_VALUE,
+    conhost_process: c.HANDLE = c.INVALID_HANDLE_VALUE,
+    resize_thread: ?std.Thread = null,
+    resize_event: c.HANDLE = c.INVALID_HANDLE_VALUE,
+    resize_lock: c.CRITICAL_SECTION = undefined,
+    resize_queue: ResizeQueue = .{},
+    running: std.atomic.Value(u8) = std.atomic.Value(u8).init(1),
```
emil-e: Why not `bool` for both of these? Would be more idiomatic in Zig.

emil-e: Also, can use the short for `.init` to not have to repeat the type.

emil-e:
@kiennq On it! Will get back with the review as soon as I can.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+
+    while (self.state.running.load(.acquire) != 0) {
+        const available = try peekOutputAvailable(self.state);
+        if (available == 0) {
+            if (shellProcessExited(self.state)) {
+                self.state.running.store(0, .release);
+                signalResizeWorker(self.state);
+                return 0;
+            }
+            _ = c.WaitForSingleObject(self.state.shell_process, 10);
```
emil-e: Why a timeout of 10 ms? Does it need a timeout if it exited already as checked above?

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+            continue;
+        }
+
+        var bytes_read: c.DWORD = 0;
+        const read_len: c.DWORD = @intCast(@min(@min(buf.len, READ_BUFFER_SIZE), available));
```
emil-e: Zig `@min` can take more than two values, so no need for multiple calls to `@min`.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+                c.ERROR_OPERATION_ABORTED, c.ERROR_BROKEN_PIPE, c.ERROR_INVALID_HANDLE => {
+                    self.state.running.store(0, .release);
+                    signalResizeWorker(self.state);
+                    return 0;
+                },
+                else => return error.ReadFailed,
+            }
+        }
+        if (bytes_read == 0) {
+            self.state.running.store(0, .release);
```
emil-e: This shutdown sequence is repeated several times. Maybe make it a separat function to ensure it always happens correctly?

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+        state.resize_thread = null;
+    }
+
+    closeConPtyHandles(state);
+    if (state.conhost_process != c.INVALID_HANDLE_VALUE) {
+        terminateProcessHandle(&state.conhost_process);
+    }
+    if (state.resize_event != c.INVALID_HANDLE_VALUE) {
+        _ = c.CloseHandle(state.resize_event);
+        state.resize_event = c.INVALID_HANDLE_VALUE;
```
emil-e: No need to ensure idempotency of deinits, so no need to reset members in this method.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    return try std.unicode.wtf8ToWtf16LeAllocZ(allocator, buf.items);
+}
+
+fn windowsPathToWtf16LeAllocZ(allocator: Allocator, path: []const u8) ![:0]u16 {
+    const normalized = try windowsPathSeparatorsAlloc(allocator, path);
+    defer allocator.free(normalized);
+    return try std.unicode.wtf8ToWtf16LeAllocZ(allocator, normalized);
+}
+
+fn windowsPathSeparatorsAlloc(allocator: Allocator, path: []const u8) ![]u8 {
```
emil-e: This can just use static memory, it doesn't need to allocate.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    const normalized = try allocator.dupe(u8, path);
+    for (normalized) |*ch| {
+        if (ch.* == '/') ch.* = '\\';
+    }
+    return normalized;
+}
+
+fn buildEnvironmentBlock(allocator: Allocator, env_map: *const std.process.EnvMap) ![]u16 {
```
emil-e: The signature can be changed to require an arena allocator, in which case it doesn't need to free anything at all within this function, simplifying the code.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    var entries = std.ArrayList([]const u16).empty;
+    defer entries.deinit(allocator);
+
+    var it = env_map.iterator();
+    while (it.next()) |pair| {
+        if (entryKeyAlreadyPresent(entries.items, pair.key_ptr.*)) continue;
```
emil-e: I don't think there can be duplicates in the EnvMap, or am I wrong? So `entryKeyAlreadyPresent` might not be needed at all.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    }
+    return true;
+}
+
+fn resizeThread(state: *State) void {
+    while (state.running.load(.acquire) != 0) {
+        if (c.WaitForSingleObject(state.resize_event, c.INFINITE) != c.WAIT_OBJECT_0) break;
+        if (state.running.load(.acquire) == 0) break;
+
+        while (takeResizeRequest(state)) |request| {
```
emil-e: Does this need to be a queue? Can't we just only keep the latest size all the time?

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+    }
+}
+
+fn resizePseudoConsole(state: *State, request: ResizeRequest) bool {
+    if (state.hpc == null) return false;
+    const size = c.COORD{
+        .X = @intCast(request.cols),
+        .Y = @intCast(request.rows),
+    };
+    return resize_pseudo_console.?(state.hpc, size) >= 0;
```
emil-e: Is resizing not thread safe?

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,831 @@
...
+
+    const pid = createdConhostPid(
+        existing_conhosts,
+        current_conhosts,
+        c.GetCurrentProcessId(),
+    ) orelse return c.INVALID_HANDLE_VALUE;
+    return c.OpenProcess(c.PROCESS_TERMINATE, c.FALSE, pid) orelse c.INVALID_HANDLE_VALUE;
+}
+
+fn createdConhostPid(
```
emil-e: Could you explain the conhost.exe process dance? I'm not sure I understand how all of this works and why it's needed.

emil-e (changes requested):
I think the architecture is exactly the right shape. Apologies for the many comments, many are just questions and most are just nits.

Really excited to be able to support WIndows officially, many people have been waiting for this.

src/ConPtyProcess.zig
kiennq: Yes, I provided emacs builds with ucrt :)
https://github.com/kiennq/emacs-build/releases

emil-e: Cool! Then we should indeed auto detect!

src/ConPtyProcess.zig
kiennq: The old code is based on the previous version where we bound to Elisp code directly and need to do resizing queue   and coalescing to prevent hang. The new structure doesn't require that anymore. I've dispatched ResizePseudoConsole directly

src/ConPtyProcess.zig
kiennq: I used to have issues with conhost is not terminated realiably and the create only return conhost handle, not process handle so the trick is to take process snapshot before and after to get the conhost process handle and terminate ot manually. For now I've changed back to just use close pseudo console. We can add back this process detecting logic later if needed.

src/ConPtyProcess.zig
kiennq: Dispatch size resize immediately now

src/PtyProcess.zig
kiennq: Removed

lisp/ghostel.el
kiennq: This is based on the old rebase. Fixed now

src/ConPtyProcess.zig
kiennq: I've added comment for this

src/ConPtyProcess.zig
kiennq: fixed

src/ConPtyProcess.zig
kiennq: added comment. This timeout is for pooling in case the process has not exited

emil-e:
@kiennq Whenever you feel ready for another review pass, just ping me. Or if you just want to discuss something, or want any help with anything.

CI has been flaky and janky recently, if you rebase then a lot of those things should work a lot better.

kiennq:
> @kiennq Whenever you feel ready for another review pass, just ping me. Or if you just want to discuss something, or want any help with anything.
> 
> CI has been flaky and janky recently, if you rebase then a lot of those things should work a lot better.

Thanks, @emil-e, can you please help to review again? I've addressed most of the comments either via code comments or inline reply.

.github/workflows/ci.yml
emil-e, line 231: How is the test suite running for the Windows backend right now? I'm okay with following that up in a later PR, or working on that part myself, but if you know they already run correctly you could add a case to the matrix. But I would guess some of them need to be tagged as platform specific already?

src/NativeProcess.zig
```diff
@@ -1,24 +1,27 @@
...
+
 const Self = @This();
 
 const log = std.log.scoped(.NativeProcessHandler);
+pub const ChannelFd = EventWriter.Fd;
+pub const ProcessParams = Backend.ProcessParams;
 
-process: PtyProcess,
-event_pipe: posix.fd_t = -1,
+process: Backend,
```
emil-e: I think `Backend` makes sense, but the member should then be called `backend` too.

src/PosixPtyProcess.zig
```diff
@@ -159,34 +212,106 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+        _ = posix.write(posix.STDERR_FILENO, @errorName(err)) catch 0;
     };
 
     const err = posix.execvpeZ(params.file, args, env);
-    // The above never returns on success, if we're here it means we failed
-    _ = posix.write(posix.STDERR_FILENO, "Failed to start subprocess: ") catch {};
-    _ = posix.write(posix.STDERR_FILENO, @errorName(err)) catch {};
+    // The above never returns on success, if we're here it means we failed.
+    _ = posix.write(posix.STDERR_FILENO, "Failed to start subprocess: ") catch 0;
+    _ = posix.write(posix.STDERR_FILENO, @errorName(err)) catch 0;
```
emil-e, line 221: Curious why `catch {}` change to `catch 0`? The latter seems more correct though.

src/PosixPtyProcess.zig
```diff
@@ -159,34 +212,106 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+
+pub fn write(self: *Self, data: []const u8) !void {
+    try self.pty.write(data);
+}
+
+pub fn drain(self: *Self, stream: anytype) !bool {
+    var drained = false;
+    var buf: [64 * 1024]u8 = undefined;
+
+    while (true) {
```
emil-e, line 241: Why a while loop? It's only ever run once, given the `break` at the end.

src/PosixPtyProcess.zig
```diff
@@ -159,34 +212,106 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+            },
+            .{
+                .fd = self.wake_pipe[0],
+                .events = posix.POLL.IN,
+                .revents = undefined,
+            },
+        };
+        _ = try posix.poll(&pollfds, -1);
+
+        if (pollfds[1].revents != 0) return drained;
```
emil-e: This can never be true, so I'd move the declaration of `drained` down to above the loop below and just return false here.

src/PosixPtyProcess.zig
```diff
@@ -159,34 +212,106 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+        break;
+    }
+
+    while (true) {
+        const len = posix.read(self.pty.primary_fd, buf[0..]) catch |err| switch (err) {
+            error.WouldBlock => return drained,
+            error.NotOpenForReading, error.InputOutput => return drained,
+            else => return err,
+        };
+        if (len != 0) {
```
emil-e: These two if cases are inversions of each other and would do better as if/else, without the `continue`.

src/GhostelTerm.zig
```diff
@@ -253,6 +255,9 @@ pub fn resize(self: *Self, cols: u16, rows: u16, cell_w: u16, cell_h: u16) !void
     self.lockTerm();
     try self.renderer.resize(cols, rows, cell_w, cell_h);
     self.unlockTerm();
+    if (self.process) |process| {
+        try process.resizePty(cols, rows);
```
emil-e: Resizing should not be necessary here. `Renderer` owns resizing and commits it on redraw. The redraw function detects it and resizes the PTY as well. This would perform resizing before it's actually confirmed.

src/ConPtyProcess.zig
emil-e: Nice. But you can ditch the member entirely. You can just use `cached_notify_crt` directly, but then maybe just call it `notify_crt`.

src/GhostelTerm.zig
```diff
@@ -86,7 +88,7 @@ pub fn redraw(self: *Self, force_full: bool) !void {
 
     if (self.isProcessLive() and !std.meta.eql(pre_size, post_size)) {
         if (self.process) |proc| {
-            try proc.process.pty.resize(post_size[0], post_size[1]);
+            try proc.process.resize(post_size[0], post_size[1]);
```
emil-e: This should use the new `resizePty` wrapper.

.github/workflows/ci.yml
```diff
@@ -160,6 +160,10 @@ jobs:
             platform: x86_64-freebsd
             suffix: .so
             target: x86_64-freebsd
+          - os: windows-latest
+            platform: x86_64-windows
```
emil-e, line 176: Do you think it might be a good idea to have aarch64 as a PR build as well? I'm not familiar with how different x86_64 and aarch64 is on Windows.

src/PosixPtyProcess.zig
```diff
@@ -159,34 +212,106 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
 
     const pid = try posix.fork();
     if (pid != 0) {
-        // This is the parent, child started successfully
+        // This is the parent, child started successfully.
         self.pty.closeReplica();
         self.pid = pid;
+        self.wake_pipe = try posix.pipe2(.{ .CLOEXEC = true });
```
emil-e: Move the pipe creation and configuration above `fork()`, it's better to do as much as possible before. It also prevents a leak where if these fail, the child is never reaped.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,589 @@
...
+            }
+            try buf.appendNTimes('\\', backslash_count * 2);
+            try buf.append('"');
+        }
+    }
+
+    return try std.unicode.wtf8ToWtf16LeAllocZ(allocator, buf.items);
+}
+
+fn buildEnvironmentBlock(arena: Allocator, env_map: *const std.process.EnvMap) ![]u16 {
```
emil-e, line 772: If my reading is correct, the environment variables must be sorted:
https://learn.microsoft.com/en-us/windows/win32/procthread/changing-environment-variables
https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,589 @@
...
+    pid: i64 = -1,
+};
+
+state: *State,
+
+var create_pseudo_console: ?CreatePseudoConsoleFn = null;
+var resize_pseudo_console: ?ResizePseudoConsoleFn = null;
+var close_pseudo_console: ?ClosePseudoConsoleFn = null;
+
+pub const ProcessParams = struct {
```
emil-e: Since this must be the same between backends, I think it should maybe live in a separate file `ProcessParams` or something like that.

src/NativeProcess.zig
```diff
@@ -27,34 +30,46 @@ term_mutex: std.Thread.Mutex.Recursive = .init,
...
-    event_pipe: posix.fd_t,
+    event_fd: ChannelFd,
 ) !void {
-    const pipe = try posix.pipe();
-    errdefer {
-        posix.close(pipe[0]);
-        posix.close(pipe[1]);
-    }
+    var process = try Backend.init(alloc, initial_cols, initial_rows, params);
+    errdefer _ = process.deinitAndWait();
```
emil-e: Since this won't force stop, the windows version could hang forever on error since the ConPTY handles are only closed after the wait.

Is the force stop mechanism needed? Can we just always close the handles before waiting? Then maybe this will just work automatically.

emil-e (changes requested):

emil-e:
@kiennq Reviewed again. Really nice work, I think this is going to be awesome. Might seem like a lot of comments but most are just nits. These parts of the code are quite tricky with lots of concurrency and ordering hazards, so I just want to make sure to get it right, I hope you don't mind. We really appreciate all the work you put in.

.github/workflows/ci.yml
dakra: Just to add another data point, since I have a Macbook with M4, the Windows I run in a VM on there is also arm Windows.
(I don't even have Emacs installed there (yet), so if it's not possible or is difficult, that shouldn't be a showstopper or anything)

src/PosixPtyProcess.zig
kiennq: the posix.write return usize on success, so I think return `0` here is more approriate

emil-e: I won't argue with that!

src/NativeProcess.zig
kiennq: `deinitAndWait` should close ConPTY handle before waiting

src/ConPtyProcess.zig
kiennq: sorted

.github/workflows/ci.yml
kiennq: Probably I will add this as follow up PR, there are some tests not work properly on Windows. I do have them run on my fork btw
https://github.com/kiennq/ghostel/actions/runs/28629733376/job/84904037684

kiennq:
@emil-e @dakra please help to review again

dakra:
@kiennq awesome, thanks.

I can't say much about the Zig code, but looks like there is a "windows-nt" branch missing for the module-download here https://github.com/dakra/ghostel/blob/13157cbc8f0859764c495c7e0cb90055cbaecd38/lisp/ghostel-module-install.el#L78

src/NativeProcess.zig
```diff
@@ -217,65 +209,33 @@ fn writeEvent(self: *Self, data: []const u8) !void {
...
-        _ = c.sigemptyset(&wait_sigs);
-        _ = c.sigaddset(&wait_sigs, posix.SIG.PIPE);
-        var sig: c_int = undefined;
-        _ = c.sigwait(&wait_sigs, &sig);
-    }
-}
-
-fn reapChild(process: PtyProcess, event_pipe: posix.fd_t) void {
+fn reapChild(process: Backend, event_writer: EventWriter) void {
     var proc = process;
```
emil-e: Should also be backend.

kiennq:
> @kiennq awesome, thanks.
> 
> I can't say much about the Zig code, but looks like there is a "windows-nt" branch missing for the module-download here
> 
> https://github.com/dakra/ghostel/blob/13157cbc8f0859764c495c7e0cb90055cbaecd38/lisp/ghostel-module-install.el#L78

Added, I forgot to port that from my fork

src/NativeProcess.zig
```diff
@@ -217,65 +209,33 @@ fn writeEvent(self: *Self, data: []const u8) !void {
...
+    writer.write(str) catch |err| {
+        log.warn("Failed to write native child exit event: {any}", .{err});
+    };
+    writer.close();
 }
 
 pub fn deinit(self: *Self) void {
     @atomicStore(bool, &self.quit, true, .monotonic);
-    _ = posix.write(self.wake_pipe[1], "X") catch {};
+    if (self.takeBackendAlive()) {
```
emil-e: There is a TOCTOU race here. Scenario:
- This thread reads `takeBackendAlive() == true`
- Reader loop exits, spawns reaper thread, deallocates resources
- `requestStop` runs and now accesses dellocated/closed resources. On the ConPty path, this is the entire state object, on POSIX, this is the wake_pipe.

I think the way to solve this is:
- Create a mutex `backend_mutex` or if you have a better name
- Remove `backend_alive` and instead make `backend` itself optional

Then something like:
```
// Emacs thread side
backend_mutex.lock();
if (backend) |*be| {
  be.requestStop()
}
backend_mutex.unlock(); 
reader_thread.join(); // join after unlock

// Reader thread side
backend_mutex.lock();
thread = spawnReaperThread(backend.?)
thread.detach();
backend = null;
backend_mutex.unlock();
```

This way, backend being present and requestStop are atomically handled. But also important to use the same mutex everywhere backend is accessed - `writePty`, `resizePty`, `replicaName`, `isBackendAlive` etc... all of it.

src/NativeProcess.zig
```diff
@@ -144,53 +174,15 @@ fn run(self: *Self) void {
...
-        if (len == 0) break;
-
-        self.stream.nextSlice(buf[0..len]);
-    }
-
-    if (pollfds[0].revents & posix.POLL.HUP != 0) {
-        return false;
-    }
+    var stream = LockedStream{ .process = self };
+    const drained = try self.backend.drain(&stream);
```
emil-e: On a closer look at this, I don't think that no bytes being drained actually means that we're completely done. I'd change this flag to not say whether we drained or not, but whether we got EOF or not.

At leat in the POSIX path, a read of `0` does not necessarily mean EOF and we should keep looping until it's truly dead.

src/PosixPtyProcess.zig
```diff
@@ -157,36 +205,102 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+            .revents = undefined,
+        },
+    };
+    _ = try posix.poll(&pollfds, -1);
+    if (pollfds[1].revents != 0) return false;
+
+    var drained = false;
+    while (true) {
+        const len = posix.read(self.pty.primary_fd, buf[0..]) catch |err| switch (err) {
+            error.WouldBlock => return drained,
```
emil-e: As per the comment in NativeProcess, I'd return `true` here but `false` for `NotOpenForReading` and `InputOutput`.

src/PosixPtyProcess.zig
```diff
@@ -157,36 +205,102 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+            error.NotOpenForReading, error.InputOutput => return drained,
+            else => return err,
+        };
+        if (len == 0) return drained;
```
emil-e: This should return `true` as described in other comments.

src/ConPtyProcess.zig
```diff
@@ -0,0 +1,611 @@
...
+    return self.state.pid;
+}
+
+pub fn drain(self: *Self, stream: anytype) !bool {
+    var drained = false;
+    var buf: [READ_BUFFER_SIZE]u8 = undefined;
+
+    while (self.state.running.load(.acquire)) {
+        const available = try peekOutputAvailable(self.state);
+        if (available == 0) {
```
emil-e: I'm not great with Windows stuff, but at least according to my understanding, you could miss bytes if you do it this way. Even if the process exited and this returns 0, there can still be trailing bytes arriving that we would miss by doing this. I think the correct thing to do is to keep reading until EOF or other errors. This mirrors the POSIX path, or at least how the POSIX path used to work. It broke a little bit now, as detailed by my other comments.

Microsoft docs seem to suggest doing the same, at least on newer versions of Windows:
https://learn.microsoft.com/en-us/windows/console/closepseudoconsole#remarks

src/PosixPtyProcess.zig
```diff
@@ -157,36 +205,102 @@ pub fn init(alloc: Allocator, initial_cols: u16, initial_rows: u16, params: Proc
...
+pub fn resize(self: *Self, cols: u16, rows: u16) !void {
+    try self.pty.resize(cols, rows);
+}
+
+pub fn write(self: *Self, data: []const u8) !void {
+    try self.pty.write(data);
+}
+
+pub fn drain(self: *Self, stream: anytype) !bool {
+    var buf: [64 * 1024]u8 = undefined;
```
emil-e: 64k buffer is a little excessive now that we're doing a drain instead of a buffered read. `4096` is a good size, I think.

emil-e (changes requested):

emil-e:
@kiennq These should be the last things as far as I'm concerned. Apologies for the back and forth but multithreaded code like this is tricky to reason about.

src/NativeProcess.zig
```diff
@@ -144,53 +174,15 @@ fn run(self: *Self) void {
...
+    if (!drained) return false;
 
     try self.notifyVtUpdate();
```
emil-e: One bug that I found on main and fixed there: `notifyVtUpdate` should be called even in the case that we're exiting to not miss the last redraw.

emil-e (dismissed):

src/NativeProcess.zig
kiennq: I'm not a fan of using mutex as it can still cause hang if we have to wait before `requestStop` can be invoked.
For example:
- `backend.write()` hang while writing to pty and take the lock with it
- `backend.requestStop()` is waiting and hang Emacs

I'm thinking of doing a queue that doesn't require lock at all.
All command (resize/write/....) will be pushed on the queue and the backend thread is reading and processing those. A thread safe queue or ring buffer will make that queue safe for both backend and the emacs thread to modifying at the same time. It doesn't wait so there should be no hang on Emacs.
When the stop happens, we simply handle backend to the reaper thread. The queue is owned by NativeProcess will still be queued but no backend process to process them anymore.

emil-e: The mutex itself is not what causes blocking, it's taking the mutex while contended and waiting for whatever operation the current holder performs. During normal operation, the only contention is between the Emacs thread doing PTY write/resize and the reader thread doing PTY write when writing to the handler through `nextSlice`. Any PTY writes resulting from stream parsing will be tiny, such as color queries. 

During shutdown, the only contention will be between `requestStop` and the creation of the reaper thread. The thread creation is not blocking in any meaningful sense and thus there won't be any meaningful waiting for the mutex during shutdown either.

The queue adds complexity that I don't think is warranted for the current scenario - PTY writes and PTY resizes are already thread safe in and of themselves on both platforms. Zig standard library doesn't have a thread safe queue, but even if it did, it wouldn't help resolve blocking since a queue can become full just like the PTY write buffer can become full and we can't throw away data.

That being said, I think being suspicious of mutexes is healthy, but in this case I still believe it's the right solution.

kiennq: I've added a mutex change and also superset that with the queue change I was talking about.
The queue change doesn't use any mutex as well as doesn't block Emacs (no wait on Emacs thread). Can you check if that makes sense?
@emil-e

emil-e:
We did some additional work on top of your changes to pull this all the way through. We went with checking Emacs quit flags instead of the queue to harden against hangs since it avoids accidentally dropping data.

PR #129, opened 2026-04-18
Fix evil-ghostel cursor snapping into scrollback on normal-state entry

Cianidos:
## Problem

`evil-ghostel--reset-cursor-point` uses `(cdr pos)` — the
viewport-relative row returned by `ghostel--cursor-position` — as a line
offset from `point-min`. That only works when the buffer has no
scrollback; with any scrollback accumulated, entering evil normal state
snaps point to row N of the **scrollback region** instead of row N of
the visible viewport.

Spotted when running Claude Code inside ghostel: Claude renders inline
(no alt-screen) so scrollback accumulates quickly, and every
`evil-normal-state` → `evil-insert-state` round-trip made point jump to
the top-left of the buffer instead of staying near Claude's input
prompt.

## Reproduction

1. Start any shell-heavy session in a ghostel buffer until there is ≥1
   screen of scrollback (running Claude Code, `cat` on a long file,
   etc.).
2. With `evil-ghostel-mode` on, trigger `evil-normal-state`.
3. Point snaps to the top-left of the buffer instead of the terminal
   cursor.

Re-entering insert state then moving further makes the display
"teleport" back down, because the terminal cursor is actually near the
bottom of the buffer.

## Fix

Offset the viewport row by the scrollback line count, mirroring the
placement math the native module performs in `src/render.zig`
(`gotoCharN(viewport_start_int)` followed by `forwardLine(cy)`, where
`viewport_start_int = point-min + forward-line scrollback_in_buffer`).

```elisp
(defun evil-ghostel--reset-cursor-point ()
  "..."
  (when (and ghostel--term ghostel--term-rows)
    (let ((pos (ghostel--cursor-position ghostel--term)))
      (when pos
        (let ((scrollback (max 0 (- (line-number-at-pos (point-max))
                                    ghostel--term-rows))))
          (goto-char (point-min))
          (forward-line (+ scrollback (cdr pos)))
          (move-to-column (car pos)))))))
```

Verified in a live buffer that `(evil-ghostel--reset-cursor-point)`
now lands on the same line as ghostel's own redraw cursor placement
(delta 0 against `(ghostel--redraw term t)`).

### DIFF
--- a/evil-ghostel.el
+++ b/evil-ghostel.el
@@ -43,13 +43,19 @@
 ;; ---------------------------------------------------------------------------
 
 (defun evil-ghostel--reset-cursor-point ()
-  "Move Emacs point to the terminal cursor position."
-  (when ghostel--term
+  "Move Emacs point to the terminal cursor position.
+`ghostel--cursor-position' returns row relative to the viewport
+(the last `ghostel--term-rows' lines of the buffer), so the row
+must be offset by the scrollback line count.  Mirrors the
+placement math the native module performs in `src/render.zig'."
+  (when (and ghostel--term ghostel--term-rows)
     (let ((pos (ghostel--cursor-position ghostel--term)))
       (when pos
-        (goto-char (point-min))
-        (forward-line (cdr pos))
-        (move-to-column (car pos))))))
+        (let ((scrollback (max 0 (- (line-number-at-pos (point-max))
+                                    ghostel--term-rows))))
+          (goto-char (point-min))
+          (forward-line (+ scrollback (cdr pos)))
+          (move-to-column (car pos)))))))
 
 (defun evil-ghostel--cursor-to-point ()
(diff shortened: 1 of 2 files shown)


dakra:
Thanks. As I don't use evil, any feedback there is welcome.

Can you add a test as well?

Cianidos:
Added `evil-ghostel-test-reset-cursor-point-with-scrollback` — writes 12 rows into a 5-row viewport (7 scroll off), then asserts the reset lands on the viewport line containing the terminal cursor, not on scrollback row N. Verified it fails against `upstream/main` (lands on `row-04` = line 5) and passes on this branch.

Also wired `ghostel--term-rows` through the test helpers — `ghostel--resize` sets it in production but direct `ghostel--new` calls in the tests don't. All 31 tests in `test/evil-ghostel-test.el` pass.

dakra:
Thanks 🙏

PR #131, opened 2026-04-18
ghostel-mode: preserve per-cell face props under font-lock

Cianidos:
## Problem

Ghostel's native renderer writes inline `face` text-properties
(`:foreground` / `:background` per cell) via `applyStyle` in
`src/render.zig`. If `font-lock-mode` ends up active in a ghostel
buffer, JIT-lock's fontify pass calls `font-lock-unfontify-region`
which strips *all* `face` props in the region — output then renders in
the buffer's default fg only.

`ghostel-mode` already runs `(font-lock-mode -1)` in its body, but
that executes *before* `after-change-major-mode-hook`, and any
globalised minor mode that force-enables font-lock later (Doom Emacs
sets `font-lock-defaults = (nil t)` globally, which makes
`turn-on-font-lock-if-desired` activate font-lock in modes that have
no keywords) flips it back on. A subsequent redraw or JIT-lock pass
then wipes the face props.

## Reproduction

In a config that forces font-lock on (Doom, or
`(setq-default font-lock-defaults '(nil t))`), spawn a ghostel buffer
running any color-heavy program (Claude Code, `btop`, colored `ls`).
Output renders monochrome: `(get-text-property POINT 'face)` returns
`nil` across most of the buffer despite the module having written
inline foreground/background properties.

Instrumentation (count of colored text-property runs) confirms the
cycle on a live ghostel buffer:

- colored runs before fontify: `2431`
- after `(font-lock-mode 1)` + `(font-lock-ensure ...)`: `0` (all
  stripped)

## Fix

Neutralise the unfontify pass via a buffer-local override in
`ghostel-mode`:

```elisp
(setq-local font-lock-unfontify-region-function #'ignore)
```

With this, `font-lock-mode` can remain active (or be toggled on by
user configuration) without ghostel's face props being wiped.
`ghostel-mode` has no font-lock keywords, so the fontify side of the
equation has nothing to add — skipping unfontify is exactly what
preserves the module's output.

Verified with the same stress pass after the fix:

- colored runs before: `91`
- after `(font-lock-mode 1)` + `(font-lock-ensure ...)`: `91`
  (preserved)

Noticed while running Claude Code inside ghostel under Doom Emacs;
companion to the evil-ghostel viewport fix in #129.

### DIFF
--- a/evil-ghostel.el
+++ b/evil-ghostel.el
@@ -45,7 +45,7 @@
 (defun evil-ghostel--reset-cursor-point ()
   "Move Emacs point to the terminal cursor position.
 `ghostel--cursor-position' returns row relative to the viewport
-(the last `ghostel--term-rows' lines of the buffer), so the row
+\(the last `ghostel--term-rows' lines of the buffer), so the row
 must be offset by the scrollback line count.  Mirrors the
 placement math the native module performs in `src/render.zig'."
   (when (and ghostel--term ghostel--term-rows)
--- a/ghostel-compile.el
+++ b/ghostel-compile.el
@@ -171,7 +171,15 @@ to a ghostel terminal."
   (setq-local next-error-function #'compilation-next-error-function)
   ;; Make sure point lands at the top after a successful recompile (and
   ;; that future input doesn't inherit ghostel's terminal-style behaviour).
-  (setq-local window-point-insertion-type nil))
+  (setq-local window-point-insertion-type nil)
+  ;; The buffer text inherited from the ghostel run carries per-cell `face'
+  ;; text-properties written by the native module.  `compilation-mode'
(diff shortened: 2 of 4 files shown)


Cianidos:
Added `ghostel-test-face-props-survive-font-lock`: writes an SGR-coloured run, renders, then force-enables `font-lock-mode` with `(setq-local font-lock-defaults '(nil t))` (mirroring how Doom Emacs sets it globally) and calls `font-lock-ensure`. Asserts the face's `:foreground` survives unchanged.

Verified it fails against `upstream/main` (face is nil after fontify — unfontify stripped it) and passes on this branch. Full `make test` suite (124 tests) passes.

dakra:
Thanks. 

jFYI, The same workaround had to be added for `ghostel-compile` as well. I added it to your commit

PR #135, opened 2026-04-18
evil-ghostel: follow terminal cursor in emacs-state

Cianidos:
## Problem

`evil-ghostel--around-redraw` freezes point in every evil state except
`insert`, so entering `emacs-state` (the vanilla-Emacs escape hatch,
typically reached via `C-z`) leaves the cursor stuck wherever it was
when the state changed while the TUI keeps redrawing elsewhere. The
TUI's own rendered cursor (e.g. Claude Code's white prompt block)
moves, but Emacs's point doesn't, so cursor-sensitive operations and
on-screen feedback go out of sync.

Insert-state is the "follow cursor" state today — and emacs-state has
the same expectation from the user's perspective: it means "turn evil
off and behave like vanilla", which in a terminal context should mean
"follow the terminal cursor".

## Reproduction

In any ghostel buffer with `evil-ghostel-mode` active, press `C-z`
to enter emacs-state. The Emacs cursor stays frozen while the
terminal keeps updating; moving the point manually is the only way
to "unstick" it. Switching back to insert-state re-enables the sync.

## Fix

- Broaden the `evil-ghostel--around-redraw` freeze-exclusion from
  `(not (eq evil-state 'insert))` to
  `(not (memq evil-state '(insert emacs)))` so redraws move point to
  the terminal cursor in both states.
- Add `evil-ghostel--insert-state-entry` to `evil-emacs-state-entry-hook`
  so switching into emacs-state syncs point immediately, matching the
  behaviour already in place for insert-state.

Includes a regression test
(`evil-ghostel-test-redraw-moves-point-emacs-state`) that enters
emacs-state, moves point away from the terminal cursor, forces a
redraw, and asserts point snaps back to the cursor column. Verified
to fail against `upstream/main` and pass on this branch.

### DIFF
--- a/evil-ghostel.el
+++ b/evil-ghostel.el
@@ -80,9 +80,13 @@ placement math the native module performs in `src/render.zig'."
   "Preserve Emacs point during redraws in evil normal state.
 ORIG-FN is the advised `ghostel--redraw' called with TERM and FULL.
 Without this, the ~30fps redraw timer would snap point back to
-the terminal cursor, undoing any evil `normal-mode' navigation."
+the terminal cursor, undoing any evil `normal-mode' navigation.
+`emacs-state' is evil's vanilla-Emacs escape hatch; point should
+follow the terminal cursor there just like it does in insert-state,
+otherwise the cursor gets stuck wherever it was on state entry while
+the TUI keeps redrawing elsewhere."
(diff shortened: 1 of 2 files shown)


dakra:
Thanks

PR #151, opened 2026-04-20
Suppress spurious SIGWINCH on minibuffer-induced window shrink

emil-e:
When the minibuffer opens (e.g. `M-x`, completion UIs like vertico/consult), Emacs shrinks the window displaying the terminal to make room. This triggers `window-configuration-change-hook` → `window--adjust-process-windows`, which calls `set-process-window-size` and delivers SIGWINCH to the shell process. When the minibuffer closes, the window grows back, sending a second SIGWINCH.

This causes visible problems in:

- **Fish shell** — repaints the entire prompt on SIGWINCH, causing it to disappear and reappear
- **Complex TUI applications** (e.g. Claude Code) — triggers a full-screen redraw for what the user perceives as a no-op

Normal (non-process) buffers simply get cropped when the minibuffer steals space — the bottom lines are hidden and reappear when the minibuffer closes. No signals, no redraws.

### Solution

When the minibuffer is active and only the height shrank (width unchanged), skip the PTY resize and return `nil` from the `adjust-window-size-function`. This suppresses `set-process-window-size` (no SIGWINCH). The Emacs window naturally hides the bottom rows. When the minibuffer closes and `window-configuration-change-hook` fires again, the size matches the committed PTY size, so again no signal is sent.

A new option `ghostel-crop-minibuffer-resize` (default `t`) controls this behavior.

Also tracks `ghostel--term-cols` alongside `ghostel--term-rows` to detect and skip no-op resize calls (e.g. when a window merely moves without changing dimensions).

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,6 +1,6 @@
 .{
     .name = .ghostel,
-    .version = "0.16.1",
+    .version = "0.16.2",
     .paths = .{""},
     .fingerprint = 0x5a44bdd1198a0f4b,
     .dependencies = .{
--- a/evil-ghostel.el
+++ b/evil-ghostel.el
@@ -4,7 +4,7 @@
 
 ;; Author: Daniel Kraus <daniel@kraus.my>
 ;; URL: https://github.com/dakra/ghostel
-;; Version: 0.16.1
+;; Version: 0.16.2
 ;; Package-Requires: ((emacs "28.1") (evil "1.0") (ghostel "0.8.0"))
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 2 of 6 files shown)


emil-e:
Would this crop something like OpenCode? OpenCode is "full screen" (and probably then alt screen). My original reason for this was to avoid resizing windows that are non-instant in their resizing - and this includes OpenCode. The reason being that it's kind of distracting when you're just running a quick command using M-x to have a terminal app briefly glitch around. Not really talking about redraw issues like before but rather just the inherent delay of complex apps.

If you don't want to crop alt screen apps, maybe it could be configurable. Tell me what you think.

Issue #146, opened 2026-04-19
PR #139 broke cursor in normal terminal usage


emil-e:
#139 causes issues with regular non-TUI terminal usage that is very easy to reproduce:
- Type something at a prompt - make sure there are no characters after the cursor

*Expectation:* Cursor gets placed after last character in command
*Actual:* Cursor gets place _on_ last character in command, both when it actually is on it and when it is past it

It's easy to see why this would happen and I feel rather dumb not predicting that it would.

emil-e:
Hmm... now I cannot reproduce it anymore. I wonder if it was a temporary glitch.

dakra:
Hmm, I've also not seen this bug yet.
You're not using evil-ghostel or something that does point calculation in elisp, right?

dakra:
I found the bug while playing with the terminfo ssh integration.
Fix incoming..

Issue #138, opened 2026-04-19
ghostel--anchor-window + scroll-conservatively: cursor-past-last-row shifts viewport up by one row


emil-e:
## Summary

When a TUI's cursor ends up in the pending-wrap state on the last visible row (i.e. col 81 of the last row after writing column 80), ghostel sets `point` to that position, which equals `window-end`. Emacs redisplay then treats `point` as off-screen and — with `scroll-conservatively` = 101 (set by `ghostel-mode`) — nudges `window-start` up by one row to make point visible. Visible symptoms: the TUI's display shifts up one line and the block cursor disappears (no cell at `window-end` to render on).

## Reproduction

1. Open a ghostel buffer with a TUI that paints its prompt/status on the last row and emits something in reaction to focus events — Claude Code is a reliable reproducer because it repaints its "unfocused" indicator on focus-out.
2. Make the prompt wide enough to fill the last row (one row of prompt isn't enough — the cursor lands somewhere in the middle of the screen, not at EOL of the last row).
3. Switch focus away from Emacs (click another app / non-Emacs window).

**Observed:** the TUI's display shifts up by one row; block cursor disappears. On refocus, a slight delay, then it shifts back.

**Does not reproduce in Ghostty proper** — Ghostty renders the pending-wrap cursor at the last column of the last row, not "below the viewport".

## Diagnosis

Instrumented `after-focus-change-function` with a trace of `window-start`, `window-point`, `window-end`, `point`, and `ghostel--last-anchor-position` for each ghostel window. Representative entries:

```
unfocus: ws=13546 anchor=13546 wp=16897 we=17046 pt=16897
refocus: ws=13547 anchor=13546 wp=17046 we=17046 pt=17046
```

Between the two focus events, `ghostel--delayed-redraw` ran (driven by the TUI's response to focus-out bytes). Inside `ghostel--anchor-window`, `pt` came in at `point-max` = `window-end`. `(set-window-start win vs t)` correctly pinned to the viewport anchor, but the subsequent redisplay cycle saw `(window-point win) = window-end`, classified it as off-screen, and `scroll-conservatively` shifted `window-start` up by one row.

## Root cause

[...]

PR #158, opened 2026-04-20
Anchor window-point clamp via pos-visible-in-window-p

### LINKED ISSUE #157: ghostel--anchor-window: viewport scrolls on focus-out when TUI parks cursor at bottom-left (pt = point-max, no pending-wrap)
## Summary

The pending-wrap-narrowed clamp added in `ad8536e` ("Narrow window-point clamp to pending-wrap only") fixes #146 but reopens a variant of #138 for TUIs that move their cursor to the bottom of the screen via CUP rather than via writing-then-wrapping. In that case `pt = point-max` and the cursor sits on the last viewport row, but `ghostel--cursor-pending-wrap-p` returns nil — so the clamp doesn't fire and Emacs scrolls `window-start` by one row to make `pt` "visible," fighting the viewport pin.

## Reproduction

- Run an interactive TUI in a ghostel buffer that sends a CUP move to `(0, last-row)` on focus loss. Anthropic's Claude Code CLI does this; likely others do too.
- Display the buffer in a window, focus a different Emacs window, then refocus.
- The viewport shifts down by one line on focus-out and back up on focus-in.

## Evidence

Captured by advising `ghostel--anchor-window` and `ghostel--focus-change` (logging `pt`, `point-max`, terminal cursor `(col . row)`, `ghostel--cursor-pending-wrap-p`, `window-start` before/after):

- **Focus-out anchor:** `pt = point-max`, `cursor-pos = (0 . 41)` (last row, col 0), `pending-wrap = nil`. After the anchor returns, Emacs redisplay shifts `window-start` by one terminal row.
- **Focus-in anchor:** `pt` is back mid-buffer, anchor restores `window-start`.

So #138's underlying scroll behavior is back; pending-wrap is just no longer the reliable predicate for it.

## Why pending-wrap is too narrow

`pending-wrap` answers "did writing past the row boundary leave the cursor in a wrap-on-next-write state?" That's one way to land at `pt = point-max` on the last visible row, but not the only way. CUP placements to `(0, last-row)` reach the same position without setting pending-wrap.

The real precondition for the bug is "Emacs considers `pt = point-max` off-screen given the current `window-start`" — which is exactly what `pos-visible-in-window-p` answers.

## Proposed fix

Draft PR: see linked PR below. Replaces the pending-wrap predicate with a direct visibility check: clamp iff Emacs would otherwise need to scroll to bring `pt` into view.

### Coverage

| Scenario | `pt = pmax` | `pos-visible-in-window-p` | Clamp? | Correct? |
|---|---|---|---|---|
| #138 pending-wrap, last row | yes | nil | yes | yes |
| #146 shell prompt mid-window | yes | t | no | yes |
| Shell prompt on last row, mid-col | yes | t | no | yes |
| Shell prompt last row + pending-wrap | yes | nil | yes | yes (visual cursor is at last cell anyway) |
| CUP-park at `(0, last-row)` (this report) | yes | nil | yes | yes |
| Mid-buffer pt | no | — | no | yes |

`pending-wrap` becomes a special case of "pt off-screen", so the #138 fix is preserved and #146's mid-window typing case continues to render correctly.

## Tested locally

Patched my install with the proposed change. Original symptom (focus-shift on Claude Code buffer) is gone. Normal shell prompt typing still renders the block cursor after the last character, including past the right margin.

## Notes

- `ghostel--cursor-pending-wrap-p` is no longer needed by `ghostel--anchor-window`. Could be kept for other consumers / tests.
- `pos-visible-in-window-p` consults the just-pinned `window-start`, so the answer reflects the state we want it to.
- Worth adding a regression test that places the cursor at `(0, body-height-1)` via CUP (not pending-wrap) with `pt = point-max` and asserts `wp` is clamped.

emil-e:
Fixes #157.

## Summary

- Replace the `ghostel--cursor-pending-wrap-p` predicate in `ghostel--anchor-window` with `(not (pos-visible-in-window-p pt win))`.
- `pending-wrap` is one way to land at `pt = point-max` on a position Emacs treats as off-screen, but not the only one — CUP placements to `(0, last-row)` reach the same state without setting pending-wrap, which lets the redisplay-induced `window-start` scroll re-emerge (variant of #138).
- The direct visibility check captures the actual precondition: clamp iff Emacs would otherwise scroll `window-start` to bring `pt` into view. `pending-wrap` becomes a special case of this, so the #138 fix is preserved; mid-window typing at a shell prompt continues to render the block cursor after the last character (#146).

See #157 for reproduction, evidence (focus-change/anchor-window log on Claude Code as the example TUI), and case-by-case coverage table.

## Test plan

- [x] Manual: original symptom (focus-shift on TUI buffer that parks cursor at bottom-left) is gone.
- [x] Manual: normal shell prompt typing renders the block cursor after the last character, including past the right margin.
- [ ] Add a regression test: place cursor at `(0, body-height - 1)` via CUP (not pending-wrap) with `pt = point-max`, assert `wp` is clamped.
- [ ] Confirm existing #138 / #146 tests still pass.

Marked draft pending the regression test and a maintainer sanity check on `pos-visible-in-window-p` performance inside the redraw path (called once per anchored window per redraw — likely fine but worth confirming against any existing benchmarks).

### DIFF
--- a/ghostel.el
+++ b/ghostel.el
@@ -3052,26 +3052,27 @@ No-op when `ghostel--snap-requested' (user input overrides)."
 Also resets pixel vscroll (pixel-scroll-precision-mode may leave a
 partial offset that would clip the top line after a redraw).
 
-When the TUI cursor is in pending-wrap state on the last visible row,
-PT equals `point-max' (one past the last character).  Emacs redisplay
-then classifies it as off-screen, and `scroll-conservatively' shifts
-`window-start' up by one row to make it visible — which fights the
-viewport pin and makes the block cursor disappear.  Clamp
+When PT equals `point-max' and Emacs considers that position
+off-screen given the just-pinned `window-start', redisplay would
+scroll `window-start' to bring PT into view — fighting the viewport
+pin and making the block cursor disappear (issue #138).  Clamp
 `window-point' back by one only in that case so it sits inside the
-viewport; buffer-point is unaffected and subsequent redraws recapture
-the real cursor.  We must NOT clamp for a plain shell prompt where the
-cursor is legitimately at `point-max' after typing — doing so would
-draw the block cursor on the last character instead of after it
-\(issue #146)."
+viewport; buffer-point is unaffected and subsequent redraws
+recapture the real cursor.  When PT is already visible (e.g. a shell
+prompt mid-window where the user just typed up to `point-max'),
+leave WP alone so the block cursor renders after the last character
+\(issue #146).
+
+This catches both pending-wrap (#138) and CUP placements that land
+the cursor at `point-max' on the last viewport row, e.g. TUIs that
+park the cursor at the bottom on focus-out."
   (set-window-start win vs t)
   (set-window-vscroll win 0 t)
-  (set-window-point win (if (and (= pt (point-max))
-                                 (> pt (point-min))
-                                 ghostel--term
-                                 (ghostel--cursor-pending-wrap-p
-                                  ghostel--term))
-                            (1- pt)
-                          pt)))
+  (set-window-point win pt)
+  (when (and (= pt (point-max))
+             (> pt (point-min))
+             (not (pos-visible-in-window-p pt win)))
+    (set-window-point win (1- pt))))
 
 (defun ghostel--restore-scrollback-window (win state)
   "Restore WIN to ws/wp recorded in STATE and push STATE to scroll-positions.


emil-e:
Closing — the proposed approach (replace `pending-wrap` predicate with `pos-visible-in-window-p`) is unsound. `pos-visible-in-window-p` reflects the previous redisplay rather than the just-pinned `window-start`, and manual `count-screen-lines` substitutes have edge-case failures in both directions. See https://github.com/dakra/ghostel/issues/157#issuecomment-4283725894 for the full analysis and a suggested terminal-side direction.

PR #161, opened 2026-04-20
Expose initial evil state as defcustom

Cianidos:
## Problem

`evil-ghostel-mode` hard-codes `(evil-set-initial-state 'ghostel-mode 'insert)` inside its activation body ([evil-ghostel.el:337](https://github.com/dakra/ghostel/blob/main/evil-ghostel.el#L337)). Wired to `ghostel-mode-hook` per the README:

```elisp
(use-package evil-ghostel
  :hook (ghostel-mode . evil-ghostel-mode))
```

the call fires on **every ghostel buffer creation**, running before `evil-initialize` reads the initial state via `after-change-major-mode-hook`. User overrides (Customize, `setq`, direct `evil-set-initial-state`) survive until the next new ghostel buffer opens, then get clobbered.

## Change

- Add `evil-ghostel-default-state` defcustom. `:set` re-runs `evil-set-initial-state` so Customize / `setopt` take effect immediately. Direct calls to `evil-set-initial-state` also keep working — the registry is last-writer-wins.
- Apply the current value once at load — covers `(setq ... )` before `require`, where `defcustom` preserves the value without calling `:set`.
- Drop the hard-coded call from the activation body.

Default stays `'insert`; no behaviour change for existing users.

## Tests

Three new ERT cases in `test/evil-ghostel-test.el`:

- `default-state-load-applied` — value is registered at load.
- `default-state-custom-set-updates-registry` — Customize propagates.
- `mode-activation-preserves-initial-state` — regression guard: activation no longer clobbers the registry.

### DIFF
--- a/evil-ghostel.el
+++ b/evil-ghostel.el
@@ -28,6 +28,35 @@
 (require 'ghostel)
 
 ;; ---------------------------------------------------------------------------
+;; Customization
+;; ---------------------------------------------------------------------------
+
+(defgroup evil-ghostel nil
+  "Evil-mode integration for ghostel."
+  :group 'ghostel
+  :prefix "evil-ghostel-")
+
+(defcustom evil-ghostel-initial-state 'insert
+  "Initial evil state for new `ghostel-mode' buffers.
+Setting this option via `customize-set-variable', `setopt', or the
+Customize UI calls `evil-set-initial-state' so the change takes effect
+immediately.  Users who prefer the raw API can call
+`evil-set-initial-state' directly from their config — the registry is
+last-writer-wins."
+  :type '(choice (const :tag "Emacs" emacs)
+                 (const :tag "Insert" insert)
+                 (const :tag "Normal" normal)
+                 (symbol :tag "Other state"))
+  :set (lambda (sym val)
+         (set-default-toplevel-value sym val)
+         (evil-set-initial-state 'ghostel-mode val)))
+
+;; Apply the current value at load.  Covers the case where the user set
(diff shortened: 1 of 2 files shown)


Cianidos:
Context on why this matters to me, in case it's useful for the design discussion:

I prefer defaulting new ghostel buffers to `'emacs` because ESC then passes through to the terminal untouched — which is essential for TUI apps (htop, btop, fzf pickers, vim over ssh, etc.).

`evil-ghostel` seems designed around a normal ↔ insert flow: insert to interact with the TUI, normal to navigate the buffer. In practice I find normal ↔ emacs (toggled via `C-z`) smoother for the same flow — emacs-state behaves like a vanilla terminal, normal-state gives me the evil goodies (scrollback, search, yank).

Possible I'm using the package against its intended grain; happy to hear if there's a more canonical way. Either way, the defcustom lets both workflows coexist without config-side advice tricks.

dakra:
Thanks again ❤️ 

> evil-ghostel seems designed around a normal ↔ insert flow: insert to interact with the TUI, normal to navigate the buffer. In practice I find normal ↔ emacs (toggled via C-z) smoother for the same flow — emacs-state behaves like a vanilla terminal, normal-state gives me the evil goodies (scrollback, search, yank).

> Possible I'm using the package against its intended grain; happy to hear if there's a more canonical way. Either way, the defcustom lets both workflows coexist without config-side advice tricks.

Personally I'm not a evil user. I just added this package because a few users requested it and it's very popular.
So if something is not evil-idiomatic etc, you can just say or create a PR to fix it.

PR #162, opened 2026-04-21
Drop normal-state-entry hook — it clobbers post-operator point

Cianidos:
## Problem

`evil-ghostel--normal-state-entry` snaps point to the terminal cursor on every normal-state entry.  `evil-normal-state-entry-hook` fires on **every** `(evil-normal-state)` call — including the no-op re-entries operators and ESC-from-visual perform at completion.

Evil's `evil-change-to-previous-state` also rewinds `evil-previous-state` through `evil-previous-state-alist` when operators finish, so even a gate like `(memq evil-previous-state '(emacs insert))` lets the bad case through.  Concretely:

- Move in normal, then `yy`: at the hook fire `this-command = evil-yank`, `evil-previous-state = emacs` (rewound from `operator`).  Snap discards the motion; point lands on the original TUI cursor row instead of the yanked line.
- `v..y`: point snapped off the yank start.
- `v..<escape>`: point snapped off visual-beginning.
- `yiW` worked correctly — text-object yanks don't round-trip through normal state.

## Why it's safe to remove

The snap is redundant in the legitimate cases too:

- `emacs → normal`: the redraw advice keeps point following the TUI cursor throughout `emacs-state`, and emacs-state-exit leaves point at the TUI cursor as `normal-state` begins.
- `insert → normal`: `evil-ghostel--insert-state-entry` is registered on `emacs`/`insert` entry; point is already at the TUI cursor until ESC.

Evil's operator/visual machinery places point correctly at every relevant transition.  Remove the function and its hook registration outright.

## Tests

- Flip the activation assertion to check the hook is NOT installed (regression guard).
- Drop the removed-function's test case.

### DIFF
--- a/evil-ghostel.el
+++ b/evil-ghostel.el
@@ -144,11 +144,6 @@ In alt-screen mode, defer to the terminal's cursor style."
   "When non-nil, skip arrow-key sync in the insert-state-entry hook.
 Set by the `I'/`A' advice which send Home/End directly.")
 
-(defun evil-ghostel--normal-state-entry ()
-  "Snap Emacs point to the terminal cursor when entering normal state."
-  (when (and (derived-mode-p 'ghostel-mode) (evil-ghostel--active-p))
-    (evil-ghostel--reset-cursor-point)))
-
 (defun evil-ghostel--insert-state-entry ()
   "Sync terminal cursor to Emacs point when entering `emacs-state'.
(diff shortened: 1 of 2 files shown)


dakra:
Thanks again ❤️ 
You're my evil-ghostel maintainer ;)

PR #163, opened 2026-04-21
Preserve `mark` across native redraws

Cianidos:
## Problem

`ghostel--redraw' moves every marker in the buffer:

- Full path (`redraw' in `src/render.zig:862') calls `env.eraseBuffer()' on resize/rotation and `env.deleteRegion(viewport_start, pointMax)' in the non-partial branch — every marker snaps to the deletion start (`point-min' for `eraseBuffer').
- Partial path (`DIRTY_PARTIAL') calls `env.deleteRegion' + `env.insert' per dirty row.  Markers drift asymmetrically by insertion-type, random-walking away from the user's intended position.

`point' is the renderer's cursor during the redraw pass and the TUI cursor after; that's the contract native exposes.  `mark' has no such role — it's user state (`C-SPC', `M-w', any region command) — and currently gets collapsed to `point-min' after any redraw, so copying a region in a vanilla ghostel buffer is impossible without rereading state in elisp.

## Change

Snapshot `(marker-position (mark-marker))' at the top of `redraw' and restore it on exit via Zig `defer'.  Clamped to `point-max' in case the buffer shrank.  No-op when mark was never set in this buffer (`marker-position' returns nil).

Adds `mark-marker' / `marker-position' / `set-marker' to the symbol cache and three thin `Env' helpers in `src/emacs.zig'.  Overhead is three funcalls per redraw tick.

## What this does not cover

- `point' — load-bearing for the renderer, must not be preserved at this layer.
- Package-specific markers (e.g. `evil-visual-beginning', `evil-visual-end', isearch markers) — the native module has no knowledge of them; their owners should wrap `ghostel--redraw' with their own save/restore.  `evil-ghostel' will follow up for the evil-specific markers.

## Tests

One ERT case (`ghostel-test-redraw-preserves-mark') sets mark inside the viewport, triggers a force-full redraw, verifies mark position unchanged.  Passes alongside the existing 152 tests against the built module.

### DIFF
--- a/src/emacs.zig
+++ b/src/emacs.zig
@@ -228,6 +228,18 @@ pub const Env = struct {
         return self.call0(sym.@"point-max");
     }
 
+    pub fn markMarker(self: Env) Value {
+        return self.call0(sym.@"mark-marker");
+    }
+
+    pub fn markerPosition(self: Env, marker: Value) Value {
+        return self.call1(sym.@"marker-position", marker);
+    }
+
+    pub fn setMarker(self: Env, marker: Value, pos: Value) Value {
+        return self.call2(sym.@"set-marker", marker, pos);
+    }
+
     pub fn deleteRegion(self: Env, start: Value, end: Value) void {
         _ = self.call2(sym.@"delete-region", start, end);
     }
@@ -291,6 +303,9 @@ pub const Sym = struct {
     @"point-max": Value,
     @"delete-region": Value,
     @"char-before": Value,
+    @"mark-marker": Value,
+    @"marker-position": Value,
+    @"set-marker": Value,
 
     // Text property names
     face: Value,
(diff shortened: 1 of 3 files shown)


dakra:
as always, thanks 🙏

Cianidos:
I’m glad I could help.
I've been waiting for someone to integrate ghostty into emacs, and now I really want to use it. And here it is

dakra:
@Cianidos I had a look at your emacs config and saw add-advice for title tracking (and I saw that in other integrations already too).
So I changed it now that you can customize. See: https://github.com/dakra/ghostel/pull/167

Also if you see something where you think a setting / hook etc would be handy, don't hesitate to create an issue.
Ghostel is still very young so there is room for improvements :)

PR #166, opened 2026-04-21
Fix scrollback at wrong width after horizontal resize (#160)

### LINKED ISSUE #160: Scrollback is not resized in in Claude Code and Pi Mono
Reproduction:
- Run Claude Code or Pi in Ghostel
- Accumulate enough output so that some overflows into scrollback
- Resize window horizontally so that content reflows
- Scroll up into scrollback

Expectation: Content in scrollback has also been resized correctly
Actual: Content in scrollback stays the same width.

Ghostty proper also has some issues with this for Claude Code. However, Pi resizes flawlessly in Ghostty. If I were to guess, Claude Code has some glitches that Pi doesn't have, but Ghostel has some further glitches.

I have not debugged this further so I don't know if the issue is that the content is not reflowing correctly inside libghostty or whether this is an issue of the scrollback not being invalidated correctly.

emil-e:
**Draft — initial approach, needs architectural review before merging.**

Fixes #160. After a horizontal resize, scrollback content was staying at the old terminal width.

## Root cause

After a resize, SIGWINCH is sent to the application AFTER the resize handler returns. Until the app processes SIGWINCH, it continues writing at the old terminal width. That old-width output gets promoted into scrollback by the row-promotion optimization (which trusts buffer content without verifying it against libghostty). Additionally, when the app clears scrollback (`CSI 3 J`) and rewrites at the new width, the scrollback sync only tracks row *counts* — if the new count coincidentally matches the old count, stale content is never replaced.

Ghostty proper does not have this problem because it renders directly from libghostty's grid on every frame — there is no cached scrollback copy that can go stale.

## Fix (two layers)

1. **Flush-before-resize** (`ghostel.el`, 1 line): feed pending PTY output to libghostty at the OLD terminal width before calling `ghostty_terminal_resize`. The resize then reflows that content correctly. Handles the common case.

2. **Scrollback content verification** (`render.zig`): on every redraw with new output, byte-compare the last materialized scrollback row in the buffer against what libghostty has at that position (via `verifyScrollbackRow` → `buildRowContent`). On mismatch — whether from `CSI 3 J` + rewrite, scrollback cap rotation, or any other cause — erase and rebuild. Always correct, no timing heuristics.

Layer 2 replaces the old rotation detection (`first_scrollback_row_hash` / `computeFirstScrollbackRowHash`), which only hashed 16 cells of one row and could miss changes when the app's header happened to match. The new approach is both simpler and more robust — it handles rotation, content replacement, and resize reflow mismatches with a single mechanism.

## Review questions

- Is the per-redraw cost of `verifyScrollbackRow` acceptable? It does one viewport scroll + render state update + row read + buffer string extraction per redraw cycle with output.
- Is it safe to remove the rotation hash entirely in favor of content verification?
- The verification checks only the last materialized scrollback row. Should it also check the first row for extra robustness?


emil-e:
Superseded by #182, which takes a cleaner approach: scan for CSI 3J in vtWrite to set a rebuild_pending flag, and unify all scrollback invalidation signals into a single erase point. This avoids the content-verification approach here and also correctly handles CSI 3J followed by enough output to restore the original scrollback depth.

Issue #168, opened 2026-04-21
ghostel-project creates a new shell on each invocation


nhojb:
The `ghostel-project` docs state:

> If a buffer already exists for this project, switch to it. Otherwise create a new Ghostel buffer.

However, when I run `ghostel-project` in a project I get a new shell buffer every time. It never switches to the current shell. So you end up with "*ghostel <project-name>*", "*ghostel <project-name>*<2>" etc, etc

> GNU Emacs 30.2.50 (build 1, aarch64-apple-darwin25.3.0, NS appkit-2685.40 Version 26.3.1 (a) (Build 25D771280a)) of 2026-03-23
> ghostel 0.17.0

dakra:
Does it work with only `ghostel` and only `ghostel-project` is always creating new buffers?

How are the buffers exactly called?
When the buffer gets renamed (e.g. from claude-code) then it'll create a new buffer but maybe that's a feature and not a bug :D
(probably a bug that I should fix too but doesn't sound like the problem you're having)

nhojb:
Yeah, it's `ghostel-project`.

After running it twice I end up with two buffers named: `*ghostel <project-name>*`, `*ghostel <project-name>*<2>`

I'll take a look and create a PR 😄

nhojb:
PR #169

dakra:
Thanks. I created yesterday already a fix for the buffer renaming bug but wanted to make sure that this is the problem you're really having.

Should be fixed on main.
Thanks for the bug report (and PR).

Please report anything that's not working as expected.

nhojb:
Thanks for the update. Glad it got fixed.

PR #169, opened 2026-04-22
Pin buffer name when creating via ghostel-project

nhojb:
Fixes issue #168 where project buffer reuse would fail due to ghostel buffer renaming.


dakra:
Solved in https://github.com/dakra/ghostel/commit/465030e50a7971209d19b6b6123bbd87c6fafa09

PR #182, opened 2026-04-23
Fix scrollback correctness: rebuild_pending flag + unified erase path

### LINKED ISSUE #160: Scrollback is not resized in in Claude Code and Pi Mono

emil-e:
## Summary

Fixes a scrollback correctness bug where stale rows could appear in the Emacs buffer after `CSI 3 J` (erase-scrollback) followed by enough new output to restore the same scrollback depth. Also addresses the related issue where resize (which triggers libghostty reflow) could leave the buffer in an inconsistent state.

Closes #160.

## Root cause

The old code had two independent validity signals (`resize_pending` and `scrollback_cleared`) with inconsistent contracts, plus a surgical trim path that assumed scrollback shrinks only by top-row eviction. All three were semantically wrong for at least one scenario:

- **CSI 3J + refill**: if 10 scrollback rows exist, `\x1b[3J` is sent, then 10 new rows are written — libghostty's row count is the same, the first-row hash may match a coincidence, and the count-based trim fires 0 lines. The stale pre-clear rows survive into the next redraw.
- **Surgical trim**: assumed the bottom N rows are intact after eviction; false for reflow and for CSI 3J.
- **Two flags, two contracts**: `resize_pending` pre-zeroed scrollback state in `terminal.resize()`; `scrollback_cleared` did not. Different callers, different behavior.

### `src/terminal.zig`
- Replace `resize_pending` (and the separate `scrollback_cleared` field that no longer exists) with a single `rebuild_pending: bool`. Callers only set the flag; the redraw pass owns all state resets.
- `terminal.resize()` no longer zeroes `scrollback_in_buffer` / `first_scrollback_row_hash` directly — that's the redraw's job.

### `src/module.zig`
- After the vtWrite loop, scan the raw input slice for `\x1b[3J`. libghostty processes the sequence and clears its internal scrollback, but provides no notification callback. Setting `rebuild_pending` here ensures the next redraw erases the Emacs buffer and rebuilds from libghostty's current state.

### `src/render.zig`
- Replace four scattered erase paths (resize_pending block, rotation hash mismatch erase, surgical trim, scrollback_cleared check) with a single `scrollback_stale` local that collects all three invalidation signals (flag, hash change, count shrink) and erases once at a single point.
- The rotation hash check is skipped when `scrollback_stale` is already true — the viewport scroll it requires would be wasted work.
- Remove the surgical trim (`else if (libghostty_sb < term.scrollback_in_buffer)`) entirely. It was the semantically incorrect path described above. The count-shrink branch now just sets `scrollback_stale`; the single erase block handles the rest.

### `test/ghostel-test.el`
- Add `ghostel-test-scrollback-csi3j-then-refill`: fills 10 "before" rows, redraws, sends `\x1b[3J`, writes 5 "after" rows (restoring partial depth), redraws, and asserts that rows 0–4 ("before-00"–"before-04") are gone while rows 5–9 and "after" rows are present. This is the exact scenario that was previously undetectable.

## Supersedes

This PR supersedes #166 (draft, different approach — content verification via last-row byte compare and flush-before-resize). That approach cannot handle the CSI 3J + refill scenario either, since it relies on content comparison rather than tracking the operation itself.

### DIFF
--- a/src/module.zig
+++ b/src/module.zig
@@ -165,6 +165,7 @@ fn fnWriteInput(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value, _: ?*
     // pair split across two writes — chunk A ending with \r, chunk B
     // starting with \n — is not mis-normalized into \r\r\n.  The final
     // value is persisted back for the next call. An empty input
+
     // round-trips the flag unchanged.
     var seg_start: usize = 0;
     var prev_was_cr: bool = term.last_input_was_cr;
@@ -182,6 +183,23 @@ fn fnWriteInput(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value, _: ?*
         term.vtWrite(raw[seg_start..]);
     }
     term.last_input_was_cr = prev_was_cr;
+
+    // Detect CSI 3 J (erase scrollback). libghostty processes it and clears
+    // its scrollback, but provides no notification. Set rebuild_pending so the
+    // next redraw erases the Emacs buffer and re-fetches from libghostty.
+    // The flag is necessary because CSI 3 J followed by enough new output to
+    // restore the same scrollback depth is undetectable by count or hash alone.
+    //
+    // This is a stateless substring scan over a single write, so it misses:
+    // (1) sequences split across two writes (ESC in one, "[3J" in the next);
+    // (2) non-canonical forms — parameter padding like `\x1B[03J`, or the
+    //     8-bit C1 form `\x9B3J`.
+    // libghostty's stateful VT parser still clears its scrollback in those
+    // cases, so the `libghostty_sb < scrollback_in_buffer` shrink check in
+    // redraw() is the fallback that catches them.
+    if (std.mem.indexOf(u8, raw, "\x1B[3J") != null) {
+        term.rebuild_pending = true;
+    }
 
     // Scan for OSC sequences that libghostty-vt discards (7, 51, 52, 133).
     // One pass, dispatched by code in document order.
@@ -624,7 +642,7 @@ fn fnSetSize(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value, _: ?*any
         return env.nil();
     };
     // Reflow invalidates the materialized scrollback — terminal.resize()
-    // sets resize_pending so the next redraw() erases and rebuilds under
+    // sets rebuild_pending so the next redraw() erases and rebuilds under
     // inhibit-redisplay, avoiding a visible blank frame.
 
     return env.nil();
--- a/src/render.zig
+++ b/src/render.zig
@@ -819,7 +819,6 @@ fn insertScrollbackRange(
     return inserted;
 }
 
-
 /// Convert a terminal column to an Emacs character offset by iterating
 /// the row's cells.  Returns `true` and positions point on success;
 /// `false` if the cell data is unavailable (caller should fall back to
@@ -891,8 +890,8 @@ fn positionCursorByCell(env: emacs.Env, term: *Terminal, cx: u16, cy: u16) bool 
 /// On each call we:
 ///   1. Force libghostty's viewport to the bottom (active screen).
 ///   2. Poll `getTotalRows()` against `term.scrollback_in_buffer` to
-///      detect rows that scrolled off the top (append to buffer) or
-///      rows evicted by libghostty's scrollback cap (trim from buffer).
+///      detect rows that scrolled off the top of the viewport and promoted them
+///      to the scrollback part of the buffer
 ///   3. Render the viewport into the tail of the buffer, anchored at
 ///      the line that follows the last scrollback row.
 ///
@@ -936,17 +935,6 @@ pub fn redraw(env: emacs.Env, term: *Terminal, force_full_arg: bool) void {
         return;
     }
 
-    // ---- Deferred resize buffer reset ----------------------------------------
-    // terminal.resize() sets resize_pending instead of immediately erasing the
-    // buffer, so the old frame stays visible until this redraw replaces it
-    // (the caller binds inhibit-redisplay around us).  Erase now and force a
-    // full rebuild of scrollback + viewport.
-    if (term.resize_pending) {
-        env.eraseBuffer();
-        term.resize_pending = false;
-        force_full = true;
-    }
-
     // Resolve default colors once — used for both the scrollback append
     // path and the viewport render path.  These always succeed (the
     // render state always has resolved default colors), so batching is safe.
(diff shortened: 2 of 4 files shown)


emil-e:
Tests are broken... will get back and fix that. But I think the approach is correct.

dakra:
This looks awesome. Thank you very much.

I ran the benchmarks vs main:

| Scenario | main (ms/iter) | pr-182 (ms/iter) | Delta |
|---|---|---|---|
| **Streaming (most relevant)** | | | |
| `stream/ghostel-incr` | 15.9 ms | 25.2 ms | **+9.3 ms (+58%)** |
| `stream/ghostel-full` | 16.0 ms | 24.9 ms | **+8.9 ms (+56%)** |
| `stream/ghostel-nodetect` | 14.5 ms | 23.5 ms | **+9.0 ms (+62%)** |
| **Engine micro** | | | |
| `engine/plain` | 9.8 ms | 11.2 ms | +1.4 ms (+14%) |
| `engine/styled` | 9.7 ms | 13.8 ms | **+4.1 ms (+42%)** |
| `engine/unicode` | 7.3 ms | 8.1 ms | +0.8 ms (+11%) |
| **PTY (real-world)** | | | |
| `pty/plain/ghostel-incr` | 15.3 ms | 15.6 ms | ~flat |
| `pty/plain/ghostel-full` | 14.7 ms | 15.9 ms | +1.2 ms (+8%) |
| `pty/urls/ghostel` | 28.8 ms | 31.2 ms | +2.4 ms (+8%) |

and asked claude for analysis:

-----

Analysis

  There's a real regression, most visible in the streaming benchmark (~58%) and partially in the engine microbenchmark (~14–42%). The PTY benchmark is much less affected (~8%) because PTY throughput is dominated by process I/O and timer
  batching, not the scan.

  The culprit is almost certainly the std.mem.indexOf(u8, raw, "\x1B[3J") scan added to every vtWrite call in module.zig. It's a full linear scan over the entire input chunk on every single write. For a 1 MB stream split into 4 KB chunks
  (~256 calls), that's 256 × 4 KB = 1 MB of memscanning per MB of input — effectively doubling the memory bandwidth through the write path.

  The streaming benchmark exercises this directly (4KB chunks, many calls), which is why the regression is largest there. The PTY benchmark batches chunks via the timer loop, so far fewer but larger writes are made, masking the overhead.

  Recommendation to the PR author: the scan should be bounded. Since \x1b[3J is only 4 bytes, you could scan only the last ~16 bytes of each chunk (the sequence can only be split across a chunk boundary, not hidden deep inside one). Or scan
   a tail window of size max(len, 7) covering the overlap. That would reduce the scan cost from O(N) per write to O(1).

-----

Maybe we can make it faster by doing the recommendation? (I didn't check it).


Also here is a /review from Opus.
I don't have much time until later today, so I also didn't fact-check that and it's only smaller things, so take it as you want and sorry for the noise if something is not right. Just dumping stuff here again:

-----

### PR #182 Review: Fix scrollback correctness: rebuild_pending flag + unified erase path

#### Overview

This PR fixes a real bug (#160): after `CSI 3J` (erase-scrollback) + enough new output to restore the same scrollback depth, stale pre-clear rows survived into the next redraw because neither the row count nor the first-row hash detected a change. The PR also cleans up the scattered invalidation logic in `render.zig`.

#### Is this the right approach?

The root constraint is that **libghostty provides no callback for `CSI 3J`**, so the bridge must detect it out-of-band. Three approaches were available:

1. **Scan raw input in `vtWrite`** — what this PR does
2. **Content verification via last-row byte compare** — what PR #166 tried; rejected because it also misses the equal-count refill scenario
3. **Expose a callback in libghostty** — ideal, but out of scope

Given the constraint, scanning raw input is the pragmatic choice. The unification of the four erase paths into a single `scrollback_stale` collector is a genuine improvement regardless of the CSI 3J fix.

#### Correctness Issues

**1. Split-sequence fragility — medium risk**

`std.mem.indexOf(u8, raw, "\x1B[3J")` scans the chunk passed to a single `vtWrite` call. If a `CSI 3J` sequence is split across two PTY reads (e.g., `\x1b[3` arrives in one chunk, `J` in the next), the flag is never set.

The count-shrink check at line 197 (`libghostty_sb < term.scrollback_in_buffer`) acts as a fallback — but it only helps when scrollback shrinks. The exact scenario the PR description calls out ("undetectable by count or hash alone") — CSI 3J followed by enough output to restore the same depth — would still be missed if the sequence was also split across chunks.

In practice PTY reads are rarely byte-by-byte, so this is unlikely. But it's an unacknowledged gap worth documenting with a comment.

**2. The included test doesn't cover the stated "equal-count" scenario**

The docstring explains the hard case: 10 before rows, CSI 3J, 10 after rows → `libghostty_sb == scrollback_in_buffer`, count-shrink doesn't fire, only `rebuild_pending` saves the day. But the test uses 10 before rows + 5 after rows, producing `libghostty_sb (5) < scrollback_in_buffer (6)` — so count-shrink fires anyway. The `rebuild_pending` signal is never tested in isolation. Adding a second scenario with `dotimes (i 10)` after CSI 3J would close this gap.

#### Code Quality

**`render.zig` — positive**

The consolidation of four scattered erase paths into one `scrollback_stale` → single erase block is clearly better. The rotation hash check is correctly skipped when `scrollback_stale` is already true. Hoisting `total_rows`/`libghostty_sb` earlier to feed the count check is clean.

**Minor doc comment issue at `render.zig:95–98`**

```diff
-///      detect rows that scrolled off the top (append to buffer) or
-///      rows evicted by libghostty's scrollback cap (trim from buffer).
+///      detect rows that scrolled off the top of the viewport and promoted them
+///      to the scrollback part of the buffer
```

"detect rows…and promoted them" is a grammatical fragment. Suggest: "detect rows that scrolled off the top of the viewport (promoted to the scrollback portion of the buffer)."

**`terminal.zig` — positive**

Removing the premature state-zero in `resize()` is correct. `scrollback_in_buffer` and `first_scrollback_row_hash` should only be cleared when the buffer is actually erased (i.e., in `redraw()`), not at resize time. The rename `resize_pending` → `rebuild_pending` gives the flag more accurate semantics since it now covers two callers.

**`module.zig` — fine**

The CSI 3J comment clearly explains the rationale. No issues.

**Test file re-indentation — noise**

The diff contains ~350 lines of indentation changes to existing compile-related tests (`ghostel-test--with-compile-buffer` bodies). These are completely unrelated to the fix and make the diff significantly harder to review. They should either be reverted or sent as a separate cosmetic PR.

#### Summary

| | |
|---|---|
| **Core fix correct?** | Yes — the flag + unified erase path is the right design |
| **Test gap** | The "equal count" scenario (the hard case described in the PR) isn't covered — the test exercises count-shrink as a co-trigger |
| **Split-sequence risk** | Real but low probability; the count-shrink fallback covers most cases; worth a comment |
| **Diff noise** | ~350 lines of unrelated re-indentation obscures the real changes |

The approach is sound. I'd ask for: (1) a test variant that uses equal-depth refill to exercise `rebuild_pending` without count-shrink, (2) a note in the CSI 3J scan comment about the split-sequence limitation, and (3) the indentation noise stripped out.

emil-e:
Some of these I already found and have fixed on my end but I will take a look at the rest and update the PR.

dakra:
> Some of these I already found and have fixed on my end but I will take a look at the rest and update the PR.

Thanks. I'm mostly interested in the performance.
Even with this PR it's plenty fast enough, but I always looked out to have the best speed possible. So if we can make it faster, we should.
Also I made a note to create a libghostty feature request to expose a CSI 3J callback, that would solve our problem completely.

emil-e:
> > Some of these I already found and have fixed on my end but I will take a look at the rest and update the PR.
> 
> Thanks. I'm mostly interested in the performance. Even with this PR it's plenty fast enough, but I always looked out to have the best speed possible. So if we can make it faster, we should.

I agree performance is super important. Not at the expense of correctness, so even if it turns out that the performance regression cannot be fixed, I think it's still the right call for now - I think more people are running glitchy agents than `cat`ing 10 MB log files these days. That said, I feel like there shouldn't be that much of a cost to doing what we're doing here. There's already scans for other escape sequences: `extractOscColorQueries` + `dispatchPostWriteOscs`

> Also I made a note to create a libghostty feature request to expose a CSI 3J callback, that would solve our problem completely.

Yes, that would be the absolute best since they're already parsing it. I think it's probably a good idea in general to think what could be helpful upstream when it comes to scrollback implementation - we're doing a very non-standard thing compared to how the API is designed to work so if there were more points of introspection into some internals that would be very helpful.

emil-e:
I still haven't looked at the performance regression and some other things, will try to get to it tonight.

emil-e:
Just to have it written down, I think this analysis makes no sense:

> 
> The culprit is almost certainly the std.mem.indexOf(u8, raw, "\x1B[3J") scan added to every vtWrite call in module.zig. It's a full linear scan over the entire input chunk on every single write. For a 1 MB stream split into 4 KB chunks
> (~256 calls), that's 256 × 4 KB = 1 MB of memscanning per MB of input — effectively doubling the memory bandwidth through the write path.
> 
> The streaming benchmark exercises this directly (4KB chunks, many calls), which is why the regression is largest there. The PTY benchmark batches chunks via the timer loop, so far fewer but larger writes are made, masking the overhead.
> 
> Recommendation to the PR author: the scan should be bounded. Since \x1b[3J is only 4 bytes, you could scan only the last ~16 bytes of each chunk (the sequence can only be split across a chunk boundary, not hidden deep inside one). Or scan
> a tail window of size max(len, 7) covering the overlap. That would reduce the scan cost from O(N) per write to O(1).

It should be a rather cheap operation. We are already scanning with the OSC iterator thing which is also O(N), and this should be much cheaper as it's using `std.mem.indexOf`. But the suggestion makes no sense regardless... we need to check for it in the whole buffer. The agent must be confused and think we're scanning to detect split escape sequences across writes.

dakra:
> It should be a rather cheap operation. We are already scanning with the OSC iterator thing which is also O(N), and this should be much cheaper as it's using std.mem.indexOf. But the suggestion makes no sense regardless... we need to check for it in the whole buffer. The agent must be confused and think we're scanning to detect split escape sequences across writes.

Agree.
I did a few more tests and seems like the removel of the "surgical trim" path is the performance cuprid in the streaming case when the scrollback size is caped.

~~Can (should?) we restore it? On reflow/CSI 3J we'll rebuild now with your PR anyway, so we could leave that trim fix alone which covers the use case of 'cat'ing a large file etc.
wdyt?~~
EDIT: doesn't seem to be so simple. Maybe we should just merge this first and then think about a way how we can gain performance back (if that's even necessary)

Thanks for all your analysis so far. big help. I'm test driving this branch currently.

emil-e:
> > It should be a rather cheap operation. We are already scanning with the OSC iterator thing which is also O(N), and this should be much cheaper as it's using std.mem.indexOf. But the suggestion makes no sense regardless... we need to check for it in the whole buffer. The agent must be confused and think we're scanning to detect split escape sequences across writes.
> 
> Agree. I did a few more tests and seems like the removel of the "surgical trim" path is the performance cuprid in the streaming case when the scrollback size is caped. Can (should?) we restore it? On reflow/CSI 3J we'll rebuild now with your PR anyway, so we could leave that trim fix alone which covers the use case of 'cat'ing a large file etc. wdyt?

That change is somewhat unrelated to the original problem of the resizing and was more part of just generally improving the correctness around this code. The problem with that one is that it's not actually correct (unless I'm missing something). The fundamental issue is that the scrollback promotion mechanism uses scrollback growth as the signal for what rows to promote into scrollback - when scrollback grows, that means that viewport rows went into scrollback since that's the only place that scrollback can come from. When we're hitting the scrollback limit in Ghostel, there's no way that we can detect how many rows were actually scrolled in to scrollback - the only thing we can know is whether the oldest line changed. If it did, we know that some lines were evicted (or cleared, that is also a possibility) but how many - so the only thing we can really do is clear the scrollback.

If we had a signal from libghostty that tells us "X lines went into the scrollback" we could both make scrollback promotion more robust and not have performance fall off a cliff at scrollback capping, but I haven't really found anything. Right now, I don't see a way around the performance drop there, and I'd probably suggest that the user either increases the scrollback cap or clear the screen with `clear` if it gets there.

I could be missing something, like I said, so I'd love to hear what you think.

> Thanks for all your analysis so far. big help. I'm test driving this branch currently.

And thank you for creating this project. The reason I'm spending time on it is that I think a project like this is important for the future of Emacs. If it was a project without promise, I wouldn't bother :)

dakra:
> If we had a signal from libghostty that tells us "X lines went into the scrollback" we could both make scrollback promotion more robust and not have performance fall off a cliff at scrollback capping, but I haven't really found anything. Right now, I don't see a way around the performance drop there, 

That's actually a good idea. Over the weekend I'll make a few feature suggestions to libghostty.

> and I'd probably suggest that the user either increases the scrollback cap or clear the screen with clear if it gets there.

I would be careful with telling users to increase the scrollback. I guess it'll get triggered later then, but it's also slower and more visible. With the 5MB (and could even be 10MB) it's still nearly instantaneous for the user. Not sure if people set it to 20MB or more. (and after a long session without resize you would reach even 20MB scrollback I guess).

> And thank you for creating this project. The reason I'm spending time on it is that I think a project like this is important for the future of Emacs. If it was a project without promise, I wouldn't bother :)

Thanks again. Highly appreciated ❤️ 


Anyway, I merged this to main now. The PR is not tracked as merged because I rebased on latest main with 2 commits that where not in here and you didn't create a feature branch for it, that means GitHub doesn't let me force push on your fork.

Let's open new issues for further discussions.

PR #186, opened 2026-04-24
Return buffer from `ghostel` and `ghostel-project`

### LINKED ISSUE #185: ghostel should return the buffer
`ghostel` (and `ghostel-project`) have no meaningful return value — they return whatever `ghostel--init-buffer` happens to fall through to (a process object for new buffers, `nil` for existing ones), and neither case is documented.

This forces callers who need the buffer to use a workaround: invoking `(current-buffer)` immediately after `ghostel` inside a `save-window-excursion`, relying on `pop-to-buffer` having changed the selected buffer as a side-effect:

```elisp
(let ((buf (save-window-excursion (ghostel t) (current-buffer))))
  ...)
```

`ghostel-exec` already returns the process as a documented contract. The equivalent for `ghostel` would be returning the buffer — consistent with how most Emacs terminal commands behave (`eshell`, `vterm`, `term` all return the buffer).

**Proposed change:**

```elisp
(defun ghostel (&optional arg)
  ...
  (pop-to-buffer buffer ...)
  (ghostel--init-buffer buffer identity)
  buffer)
```

emil-e:
Both `ghostel` and `ghostel-project` now explicitly return the (live) buffer they create or switch to.  This allows callers to use the buffer programmatically without relying on `pop-to-buffer` side-effects.

### Changes

- **`ghostel`** — adds `buffer` as the final expression of the function body, so the buffer is returned instead of whatever `ghostel--init-buffer` falls through to
- **`ghostel-project`** — inherits the return value automatically since it calls `ghostel` as its final form
- **Docstrings** — both functions now document the return value; placed at the end of the docstring per Emacs convention
- **Tests** — two new tests verify the return value is a live buffer with the expected name

Closes #185

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -3477,7 +3477,8 @@ buffer creation time — see `ghostel--buffer-identity'."
 With a non-numeric prefix arg, create a new buffer.
 With a numeric prefix ARG, switch to the buffer with that number or
 create it if it doesn't exist yet.
-The name of the buffer is determined by the value of `ghostel-buffer-name'."
+The name of the buffer is determined by the value of `ghostel-buffer-name'.
+Returns the buffer."
   (interactive "P")
   (ghostel--load-module t)
   (let* ((fresh (and arg (not (numberp arg))))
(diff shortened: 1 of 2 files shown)


dakra:
Thanks

PR #188, opened 2026-04-24
Add ghostel-paste-string public API for bracketed paste

### LINKED ISSUE #187: Add public API for bracketed paste of arbitrary text
`ghostel-send-string` (added in [5453c22](https://github.com/dakra/ghostel/commit/5453c22)) is a public function that sends raw bytes to the terminal process. However, it doesn't wrap the text in bracketed paste markers (`\e[200~` / `\e[201~`), so the shell echoes and interprets the text character-by-character rather than pasting it atomically.

The internal `ghostel--paste-text` does exactly this — it sends text with bracketed paste markers when terminal mode 2004 is enabled — but being prefixed with `--`, it's not intended for external callers.

## Use case

Third-party code needs to inject arbitrary strings into the ghostel terminal in a way that the shell treats as a paste, not as typed keystrokes. This is important for reliability: without bracketed paste, multi-character input can be interleaved with shell prompts, misinterpreted by readline, or cause timing issues.

## Proposal

Expose a public function, e.g.:

```elisp
(defun ghostel-paste-string (string)
  "Send STRING to the terminal using bracketed paste.
Signals a \`user-error' when called outside a ghostel buffer."
  (unless (derived-mode-p 'ghostel-mode)
    (user-error "Must be called from a ghostel buffer"))
  (ghostel--paste-text string))
```

Alternatively, add a `:bracketed` keyword argument to the existing `ghostel-send-string`:

```elisp
(defun ghostel-send-string (string &optional bracketed)
  "Send STRING to the terminal process.
When BRACKETED is non-nil, wrap STRING in bracketed paste markers
(\\e[200~ / \\e[201~) so the shell treats it as a paste rather than
typed input."
  ...)
```

Either approach would cover the same need. The standalone function (`ghostel-paste-string`) feels cleaner as it mirrors the existing `ghostel-send-string` / `ghostel-send-key` naming pattern and keeps the concerns separate.

emil-e:
Add a public function `ghostel-paste-string` that sends arbitrary text to the terminal using bracketed paste markers (ESC [200~ / ESC [201~) when the terminal supports mode 2004.

This allows third-party code to inject strings into the terminal in a way that the shell treats as an atomic paste rather than character-by-character typed keystrokes.

Closes #187

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1388,6 +1388,18 @@ Signals a `user-error' when called outside a ghostel buffer."
     (user-error "Must be called from a ghostel buffer"))
   (ghostel--send-encoded key-name (or mods "")))
 
+(defun ghostel-paste-string (string)
+  "Send STRING to the terminal using bracketed paste.
+Signals a `user-error' when called outside a ghostel buffer.
+
+Unlike `ghostel-send-string', this wraps STRING in bracketed paste
+markers (ESC [200~ / ESC [201~) when the terminal supports bracketed
+paste mode (mode 2004), so the shell treats the input as an atomic
+paste rather than character-by-character typed keystrokes."
+  (unless (derived-mode-p 'ghostel-mode)
+    (user-error "Must be called from a ghostel buffer"))
+  (ghostel--paste-text string))
+
 
 ;;; Terminal control commands (C-c prefix)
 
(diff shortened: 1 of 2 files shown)


dakra:
Merged. Thanks.

btw, your 'main' branch seem to have diverged from mine. you probably should do a `git reset --hard origin/main` or whatever the dakra/ghostel remote is.

emil-e:
@dakra Yes, I did now, sorry about that!

PR #189, opened 2026-04-24
Replace FNV-1a rotation-detection hash with direct row snapshot

emil-e:
## Summary

Replace the FNV-1a hash of the first 16 cells with a full codepoint snapshot of the entire scrollback row. This eliminates the hash collision probability and the arbitrary 16-cell blind spot, while adding zero measurable performance overhead.

## Benchmark Results

Benchmarked on Emacs 30.2, 1 MB data, 3 iterations each. Values in milliseconds per iteration.

### Real-world PTY (cat 1.0 MB through pipe)

| Scenario | Old (ms) | New (ms) | Δ |
|---|---|---|---|
| plain/incr | 10.64 | 10.57 | −0.7% |
| plain/full | 10.33 | 10.45 | +1.2% |
| plain/nodetect | 10.75 | 10.45 | −2.8% |
| urls/ghostel | 22.99 | 23.52 | +2.3% |
| urls/nodetect | 11.54 | 10.76 | −6.8% |

### Streaming (chunked, no PTY)

| Scenario | Old (ms) | New (ms) | Δ |
|---|---|---|---|
| incr | 18.85 | 18.85 | 0% |
| full | 18.69 | 18.79 | +0.5% |
| nodetect | 17.51 | 17.65 | +0.8% |

### TUI Frame Rendering

| Scenario | Old (ms) | New (ms) | Δ |
|---|---|---|---|
| 24x80 incr | 1.766 | 1.766 | 0% |
| 24x80 full | 1.797 | 1.710 | −4.8% |
| 40x120 incr | 1.752 | 1.784 | +1.8% |
| 40x120 full | 1.755 | 1.773 | +1.0% |

### Engine Micro-Benchmarks

| Scenario | Old (ms) | New (ms) | Δ |
|---|---|---|---|
| plain/incr | 8.44 | 8.45 | +0.1% |
| plain/full | 8.50 | 8.42 | −0.9% |
| styled/incr | 12.01 | 12.07 | +0.5% |
| styled/full | 12.14 | 12.09 | −0.4% |
| unicode/incr | 6.72 | 6.71 | −0.1% |
| unicode/full | 6.63 | 6.62 | −0.2% |

**Verdict: no meaningful performance difference.** Every delta is within measurement noise. The rotation detection code path is dominated by ~6 libghostty C calls. The old FNV-1a over 16 cells took nanoseconds; the new `std.mem.eql` over 512 u32 (2KB) is also nanoseconds — both vanish compared to the C call overhead.

## Correctness improvements

- **Zero collision probability.** FNV-1a over 16 codepoints has a ~1-in-2^64 collision chance. Direct comparison on the full row is strictly correct.
- **No arbitrary 16-cell blind spot.** The old code only sampled the first 16 cells. If two rotated rows shared the same first 16 cells (e.g. a long prompt where the command happens to be identical in the first 16 chars), rotation would be silently missed. The new code samples all `term.cols` cells.
- **Explicit validity flag.** Replacing `first_scrollback_row_hash != 0` with `first_scrollback_row_valid` is cleaner — the old code used zero as a sentinel for uninitialized, conflating "has scrollback" with "has been sampled."

## Memory

`[512]u32` = 2KB vs `u64` = 8 bytes. Stored in the heap-allocated `Terminal` struct — trivial for a terminal emulator.

## Changes

- `src/render.zig`: `readFirstScrollbackRow` now reads all `term.cols` cells (capped at 512) into the output buffer; comparison uses `std.mem.eql` instead of hash comparison.
- `src/terminal.zig`: `first_scrollback_row: [512]u32` + `first_scrollback_row_valid: bool` replace `first_scrollback_row_hash: u64`.
- Benchmark methodology: run old and new side-by-side with the same benchmark harness, 3 iterations each, ReleaseFast optimization.

### DIFF
--- a/src/render.zig
+++ b/src/render.zig
@@ -464,21 +464,23 @@ pub fn isRowEmptyAt(term: *Terminal, cy: u16) bool {
     return true;
 }
 
-/// Hash the first ~16 cells of libghostty's first scrollback row using
-/// FNV-1a. Returns 0 if there is no scrollback or if anything fails.
+/// Read the first scrollback row's codepoints into `out` (one per cell,
+/// up to 512 entries). Returns true on success, false if there is no
+/// scrollback or anything fails. For terminals wider than 512 columns
+/// only the first 512 cells are compared — a practical limit since
+/// 512 columns exceeds any standard display.
 ///
 /// Used to detect rotation: when libghostty's scrollback is plateaued at
 /// its byte cap, sustained writes evict the oldest row in lockstep with
 /// new rows being pushed, so `total_rows` doesn't change and the normal
-/// delta-detection sees no work to do. Sampling the first scrollback
-/// row's content lets us detect that the row at index 0 has changed
-/// underneath us.
+/// delta-detection sees no work to do. Comparing the full row lets us
+/// detect that the row at index 0 has changed underneath us.
 ///
 /// Scrolls libghostty's viewport to the top to read the row, then
-/// restores the previous viewport offset. Cheap (~6 libghostty calls);
-/// gated by the caller to only run when rotation is suspected.
-fn computeFirstScrollbackRowHash(term: *Terminal) u64 {
-    const sb = term.getScrollbar() orelse return 0;
+/// restores the previous viewport offset. Gated by the caller to only
+/// run when rotation is suspected.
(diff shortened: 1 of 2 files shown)


dakra:
Looks good. thanks again :)

PR #190, opened 2026-04-24
Defer plain-text link detection after redraw

kiennq:
## Summary
- defer plain-text URL/file detection until after redraw and coalesce repeated scans
- keep upstream file detection semantics and native OSC-8 hyperlink handling
- cover delayed scheduling, read-only safety, and sentinel cleanup with Elisp tests

## Testing
- `emacs --batch -Q -L lisp --eval "(setq byte-compile-error-on-warn t)" -f batch-byte-compile lisp/ghostel.el`
- `emacs --batch -Q --eval "(setq load-prefer-newer t native-comp-jit-compilation nil native-comp-enable-subr-trampolines nil comp-enable-subr-trampolines nil)" -L . -L lisp -l ert -l test/ghostel-test.el -f ghostel-test-run-elisp`
- `zig build -Doptimize=ReleaseFast -Dcpu=baseline`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.18.1] — 2026-04-25
+
+### Added
+- `ghostel-plain-link-detection-delay` user option (default 0.1s)
+  controls how long ghostel waits after a redraw before scanning for
+  plain-text URLs and file paths.  Set to 0 to restore the previous
+  synchronous behavior
+  ([671d3ee](https://github.com/dakra/ghostel/commit/671d3ee)).
+
+### Changed
+- Plain-text link detection is now deferred off the redraw path and
+  coalesced via a single timer, so bursts of redraws collapse into one
+  scan instead of running detection on every dirty redraw.  Native
+  OSC-8 hyperlink spans continue to be handled inside the renderer.
+  The process sentinel cancels the pending detection timer so it
+  cannot fire against a buffer that is about to be killed
+  ([671d3ee](https://github.com/dakra/ghostel/commit/671d3ee)).
+- Scrollback rotation detection now snapshots the first scrollback
+  row directly (`std.mem.eql` over all `term.cols` cells) instead of
+  hashing the first 16 cells with FNV-1a.  Removes a small collision
+  probability and the arbitrary 16-cell sample that could miss
+  rotation when two rows shared the same opening cells; the
+  cached-read optimisation that skips the end-of-redraw round trip is
+  preserved
+  ([4b1a0ba](https://github.com/dakra/ghostel/commit/4b1a0ba)).
+
 ## [0.18.0] — 2026-04-24
 
 ### Breaking
--- a/README.md
+++ b/README.md
@@ -892,24 +892,25 @@ terminal emulators: [vterm](https://github.com/akermu/emacs-libvterm) (native
 module), [eat](https://codeberg.org/akib/emacs-eat) (pure Elisp), and Emacs
 built-in `term`.
 
-The primary benchmark streams 5 MB of data through a real process pipe,
+The primary benchmark streams 1 MB of data through a real process pipe,
 matching actual terminal usage.  All backends are configured with ~1,000
 lines of scrollback (matching vterm's default).  Results on Apple M4 Max,
 Emacs 31.0.50:
 
 | Backend              | Plain ASCII | URL-heavy |
 |----------------------|------------:|----------:|
-| ghostel              |    87 MB/s  |  64 MB/s  |
-| ghostel (no detect)  |    86 MB/s  |  86 MB/s  |
-| vterm                |    35 MB/s  |  27 MB/s  |
-| eat                  |   4.7 MB/s  | 3.5 MB/s  |
-| term                 |   5.7 MB/s  | 4.7 MB/s  |
+| ghostel              |    81 MB/s  |  77 MB/s  |
+| ghostel (no detect)  |    78 MB/s  |  75 MB/s  |
+| vterm                |    34 MB/s  |  28 MB/s  |
+| eat                  |   4.9 MB/s  | 3.8 MB/s  |
+| term                 |   5.8 MB/s  | 4.9 MB/s  |
 
 Ghostel scans terminal output for URLs and file paths, making them clickable.
-The "no detect" row shows throughput with this detection disabled
-(`ghostel-enable-url-detection` / `ghostel-enable-file-detection`).  The other
-emulators do not have this feature, so their numbers are comparable to the "no
-detect" row.
+Detection runs on a coalesced timer outside the redraw hot path, so enabling
+it costs essentially nothing on the streaming throughput — the "no detect"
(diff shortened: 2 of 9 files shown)


dakra:
Thanks. Looks useful.
The CI tests fail though.

As I'm going to bed now I just paste a quick review from Claude.
Use your own judgement if you think fixing makes sense.
I'll have a closer look tomorrow.

Here:

-----

#### Code Review: PR #190 — Defer plain-text link detection after redraw

**Overview**

Moves plain-text URL/file detection off the hot redraw path. Previously `render.zig`'s `redraw` synchronously called Elisp `ghostel--detect-urls` whenever any row was dirty. This PR removes that call, introduces a 0.1s coalescing timer in Elisp, and routes detection through a new `ghostel--queue-redraw-post-processing` helper invoked from `ghostel--delayed-redraw` and `ghostel-force-redraw`. Native OSC-8 hyperlinks are unaffected.

Changes: ~45 lines in `ghostel.el` + timer-cleanup in sentinel, 1 call removed from `render.zig`, 1 symbol removed from `emacs.zig`, and 4 new Elisp tests.

**Blocking issues**

*1. Existing native test is broken — CI is red*

`ghostel-test-scrollback-preserves-url-properties` (test/ghostel-test.el:312) fails on all 4 native-test matrix jobs. It directly calls `ghostel--redraw term t` and then immediately asserts `help-echo` is set on the URL. That worked before because `render.zig` called `ghostel--detect-urls` synchronously. Now:

- `ghostel--redraw` (the C→Elisp entry) no longer triggers detection.
- `ghostel--queue-redraw-post-processing` is only called from `ghostel--delayed-redraw` and `ghostel-force-redraw`.
- Even if the test invoked the queue, it would schedule a 0.1s timer and return before the assertion runs.

The PR needs to either (a) keep a synchronous detection path when called in a way the test can trigger, or (b) update the test to drive detection via the queue + `(timer-until ...)` / manual timer firing like the new coalescing test does. Option (b) is probably cleanest. Whichever you pick, CI must go green before this merges.

*2. Plain-text scroll-off now breaks silently*

The old render-side call passed `(viewport_start, point-max)` so a URL that landed on a line getting scrolled *off the active screen this redraw* was still within the scan window (since scrollback promotion preserves buffer text, detection on the promoted region was done once before the row left the viewport). With the new timer, `ghostel--delayed-redraw` captures `vs` **after** `ghostel--redraw` runs (line 3311), then queues detection with `(vs (point-max))`. That's correct in isolation — but if a row is scrolled off between the queue call and the 0.1s timer firing, the next redraw will compute a *new, lower* `vs` and the old row is no longer in the range.

Coalescing via `min`/`max` saves us here — the earlier call left `begin = vs_old`, and the later call's `begin = vs_new ≥ vs_old` takes the `min`, keeping the lower bound. Good. But `ghostel-force-redraw` passes no args, so `begin` defaults to the *current* viewport start — not whatever viewport was active when the last scroll happened. If a user force-redraws after URL-bearing rows have scrolled past, those rows won't get detected. Minor, since force-redraw is explicit user action, but worth noting in the docstring.

**Correctness / design**

*3. Indentation shift on unchanged lines*

In `ghostel--delayed-redraw` (lines 3321-3322) the two `setq` lines changed indentation from 10 to 11 spaces without any structural change:

```
           (setq ghostel--snap-requested nil)
           (setq ghostel--windows-needing-snap nil))))))
```

Same in `test/ghostel-test.el` around line 7746 — five pre-existing list entries gained a leading space. Both look like accidental edits. Re-indent with `indent-region` and drop the noise from the diff.

*4. `inhibit-read-only t` inside `ghostel--detect-urls`*

Adding this to the function itself is correct (the timer runs outside the dynamic-binding scope of `ghostel--delayed-redraw`), but it weakens the invariant for all other callers — tests at 2061..2345 now don't need to bind it, which is fine, but a future caller that's passing a genuinely read-only buffer they don't want mutated no longer gets an error. Given the only mutation is `help-echo`/`mouse-face`/`keymap` text properties and this is a private function, acceptable. Consider a one-line docstring mention.

*5. `inhibit-modification-hooks` is no longer in effect*

The old code ran detection inside `ghostel--delayed-redraw`'s `(inhibit-modification-hooks t)` scope (line 3306). The timer fires with no such binding. `put-text-property` on these three properties doesn't normally trigger `after-change-functions`, but combined with `buffer-modified-p` — which *can* flip — any `after-change-functions` installed by the user now runs on every redraw's detection pass. Probably benign, but if you see modeline flicker or spurious modification hooks, this is where to look.

*6. File-detection performance: `seen` hash is now per-call instead of per-redraw*

The hash `seen` memoizes `file-exists-p` across the *single* scan. Under the old model, one redraw = one detect call = one hash = high cache hit on compiler-diagnostic walls of "src/foo.rs" lines. Under the new model, detect runs at most every 100ms, which is actually *better* amortization — so this is a wash or a small win. No action needed.

*7. `ghostel-plain-link-detection-delay = 0` short-circuit*

When the delay is 0, `queue-plain-link-detection` calls `run-queued-plain-link-detection` *synchronously from inside the redraw*. That reintroduces the exact problem the PR is trying to fix (blocking the redraw on detection). If users set it to 0 they probably know what they're getting, but worth noting: the old behavior was "always synchronous during redraw," the new default is "always deferred 100ms," and 0 gives back the old cost. Fine as-is.

**Minor**

- `ghostel--queue-redraw-post-processing` name is longer than it needs to be. `ghostel--queue-detection` or similar would read easier at call sites. Keep the public name if you plan to add other post-processing hooks here.
- The `(<= begin end)` guards in both `queue-plain-link-detection` and `run-queued-plain-link-detection` are redundant with each other — the min/max logic makes it impossible for a queued range to be inverted. Not worth changing.
- Test `ghostel-test-detect-urls-allows-read-only-buffers` has inconsistent indentation on its last three `should` forms (indented one column more than the rest of the `let`). Cosmetic.

**Test coverage**

The 4 new Elisp tests exercise the right paths: defer, coalesce, read-only, sentinel cancel. The coalesce test stubs `run-with-timer`, which is the right call — fewer timing flakes in CI. Missing: a test for `ghostel-plain-link-detection-delay = 0` to prove the synchronous branch works.

**Risk summary**

| Area | Risk |
|---|---|
| CI native test regression | **High** — must fix before merge |
| User-facing URL clickability latency | Low — 100ms imperceptible |
| Scrollback URL coverage under force-redraw | Low, edge case |
| Modification-hook semantics shift | Low, but worth watching |
| Indentation/diff noise | Trivial, clean up |

**Bottom line**

The deferral is a clear win for redraw throughput and the coalescing/cancellation logic is sound. But the PR can't land while `ghostel-test-scrollback-preserves-url-properties` is red across 4 matrix jobs — either fix the test to drive the new queue, or preserve a synchronous path for the in-process test. Plus the two unrelated indentation shifts should be reverted to keep the diff surgical.

kiennq:
The CI is fixed now

dakra:
Thanks. I fixed the other issues. Added one more test as well.
Test driving a bit and then will merge.

dakra:
Merged into main.
The benchmark numbers look great now for url-heavy output.
As expected, basically no difference now for streaming output if you have file/url-detection enabled or not :)

PR #193, opened 2026-04-26
Set list-buffers-directory in ghostel-mode and ghostel-compile-view-mode

Cianidos:
## Summary
ghostel buffers visit no file, so `list-buffers-directory` is the conventional way for them to expose their cwd to buffer-list consumers (`buffer-menu`, `ibuffer`, etc.). `eshell-mode` already does this in `esh-mode.el`; ghostel did not.

## Change
One `setq-local` in `ghostel-mode`, one in `ghostel-compile-view-mode`, both setting `list-buffers-directory` to `default-directory` at mode init.

## Tests
Two pure-elisp ert tests added under `test/ghostel-test.el` and registered in `ghostel-test--elisp-tests`. Run via:

    emacs --batch -Q -L lisp -L test -l ert -l test/ghostel-test.el \
      --eval '(ert-run-tests-batch-and-exit "list-buffers-directory")'

Both pass against the current branch.

## Compatibility
Behaviour is unchanged for any code path that does not consume `list-buffers-directory`.

### DIFF
--- a/lisp/ghostel-compile.el
+++ b/lisp/ghostel-compile.el
@@ -180,7 +180,8 @@ to a ghostel terminal."
   ;; the recorded output on the first JIT-lock pass.  Neutralise unfontify:
   ;; compilation-mode's keywords are applied once via `font-lock-ensure'
   ;; on a finalised, static buffer and don't need to be cleaned up.
-  (setq-local font-lock-unfontify-region-function #'ignore))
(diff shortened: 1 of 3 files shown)


dakra:
Thanks. I looked at different modes and term.el and eat don't set it at all while eshell only in the beginning (like you did).
but shell.el sets it on every 'cd' which imho makes most sense so I added that as well.
Write here if you think otherwise.

PR #202, opened 2026-04-28
Stabilize IME preedit anchor during redraw

yours57:
## Description

This fixes IME candidate window jumping while high-throughput terminal apps such as Codex or Claude Code are
actively redrawing.

GTK/PGTK input methods display uncommitted text through preedit overlays anchored at Emacs point. Ghostel
redraws move point to the terminal cursor after each render, which can make the IME preedit/candidate window
jump while the user is composing Chinese text.

This change captures an active `x-preedit-overlay` / `pgtk-preedit-overlay` before redraw and restores it to
the same viewport row/column after the native redraw rewrites the buffer. The normal terminal cursor/window
anchoring behavior is unchanged when no preedit overlay is active.

## Changes

- Detect active X/PGTK preedit overlays in ghostel buffers.
- Preserve the composing window’s preedit overlay and `window-point` across redraws.
- Add an exact-point path to `ghostel--anchor-window` for active IME preedit state.
- Add a regression test for redraw preserving the preedit anchor.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -3166,6 +3166,69 @@ position COL columns into the first matched line."
         (forward-line (- (1- tr)))
         (line-beginning-position)))))
 
+(defun ghostel--active-preedit-overlay ()
+  "Return the active GUI preedit overlay in the current buffer, if any.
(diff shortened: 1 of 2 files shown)


dakra:
Merged. Thanks 🙏

PR #203, opened 2026-04-28
Refactor native rendering for correctness and robustness

emil-e:
## Before

This PR refactors the native rendering code for correctness and readability. The old rendering code had several issues:

- The old code had a stale-line bug: rendered viewport lines were promoted to scrollback as-is, even if they had been modified since the last render. For example:
  1. Output `foo\r\n` to the top terminal line
  2. `[render]`
  3. Move cursor to home position
  4. Output `bar`
  5. Move cursor to bottom row
  6. Output `baz\r\n`
  7. `[render]`

  The old rendering code would now have `foo` in scrollback but it should be `bar`. The added `ghostel-test-no-stale-lines-in-scrollback` test covers this case.

- The old code relied only on scrollback length to determine scrolling, which made it impossible to detect scrolling when scrollback reaches cap and stops growing. Scrolling still happens but the old rows are evicted. This was handled by comparing row 0 to a stored copy to detect this, invalidating the entire thing when that happens.

- Since we had no means of detecting screen clearing (`CSI 3 J`), a separate scan was added on the input side to detect this specific escape sequence. However, there was no handling of the case when a sequence arrived split across two separate writes. In addition, if libghostty were to add support for other ways of clearing the scrollback, we would not be able to detect that.

## After

The new code uses a different algorithm that takes advantage of a very specific trick. At the end of every render pass, we position the terminal viewport at `max_offset - 1` so that the last row is just outside the viewport. This has two important properties:
- The viewport is no longer pinned to the active area and instead follows scrollback naturally, even when we reach the scrollback cap.
- If the scrollback is cleared, libghostty resets the viewport to the bottom (`offset + len == total`), which we detect and use as a signal to clear our materialized scrollback.

The viewport cannot be positioned at `max_offset - 1` if we don't have any scrollback. In that case, we don't need to detect scrollback clearing (it was already empty) and scrolling can be detected by simply checking the scrollback length.

A simplified version of the new algorithm is as follows:
1. If we had scrollback and the viewport position is now at the bottom, reset scrollback and force a full redraw.
2. Unpark the viewport:
   - If we had scrollback: move libghostty viewport `+1` to compensate for the `-1` position, and move Emacs point to the end of the materialized scrollback.
   - If we had no scrollback: move libghostty viewport to the top, and move Emacs point to the start of the buffer.
3. Walk from the current viewport position to the active area in viewport-sized steps, rendering each chunk. Consecutive positions overlap when the remaining range is smaller than a full viewport; overlapping rows are skipped since they were already rendered at the tail of the previous step.
4. After the walk, Emacs point and the libghostty viewport are both at the active area. Non-dirty rows in the first step (where the old active area was) are reused as-is; all other steps render fully.
5. Position viewport at `max_offset - 1` for the next render.

## Caveats
- Benchmarks show a slight increase in PTY performance, but a drop in performance in the chunked streaming tests. This is to be expected since we are no longer promoting incorrect rows. Still much faster than `vterm` but further work should go into see if we can win some performance back. Profiling shows two main culprits:
  - Reading libghostty cell properties - this is hard to get around
  - Comparing styles to separate style runs - this could potentially be optimized
  
  One future optimization could be to check if a line has any styling at all (libghostty has APIs for this) and skip cell checks entirely if it does not.

- When libghostty rotates rows out at the scrollback cap, those rows are not evicted from the Emacs buffer. This is intentional: the Emacs buffer serves as the permanent record, and libghostty's scrollback cap only bounds libghostty's memory use, not what is visible in Emacs.

- This PR also includes a change to how OSC8 links are handled. Rather than attaching the URI to separate URI runs, we can avoid the URI comparison entirely and just make "has URI" part of the style so it automatically gets treated as separating style runs. We then attach a lookup function instead of a string as `help-echo` that calls down into libghostty to fetch the URI only when hovered or clicked. This should really have been a separate PR but it was hard to separate from what has been a rather organic process. Apologies for that. If you really want it to be separate, I can try to make that happen.

- The buffer now always includes a trailing newline. This makes things more uniform (every line has a newline at the end) and makes the line math easier. It also removes the need for point-max-clamping, at least according to my tests.

### DIFF
--- a/extensions/evil-ghostel/evil-ghostel.el
+++ b/extensions/evil-ghostel/evil-ghostel.el
@@ -85,7 +85,7 @@ placement math the native module performs in `src/render.zig'."
   (when (and ghostel--term ghostel--term-rows)
     (let ((pos (ghostel--cursor-position ghostel--term)))
       (when pos
-        (let ((scrollback (max 0 (- (line-number-at-pos (point-max))
+        (let ((scrollback (max 0 (- (count-lines (point-min) (point-max))
                                     ghostel--term-rows))))
           (goto-char (point-min))
           (forward-line (+ scrollback (cdr pos)))
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -567,8 +567,6 @@ Bump this only when the Elisp code requires a newer native module
 ;; Declare native module functions for the byte compiler
 
 (declare-function ghostel--cursor-position "ghostel-module")
-(declare-function ghostel--cursor-pending-wrap-p "ghostel-module")
-(declare-function ghostel--cursor-on-empty-row-p "ghostel-module")
 (declare-function ghostel--encode-key "ghostel-module")
 (declare-function ghostel--focus-event "ghostel-module")
 (declare-function ghostel--mode-enabled "ghostel-module")
@@ -578,11 +576,11 @@ Bump this only when the Elisp code requires a newer native module
 (declare-function ghostel--mouse-event "ghostel-module")
 (declare-function ghostel--new "ghostel-module")
 (declare-function ghostel--redraw "ghostel-module" (term &optional full))
-(declare-function ghostel--scroll-bottom "ghostel-module")
 (declare-function ghostel--set-default-colors "ghostel-module")
 (declare-function ghostel--set-palette "ghostel-module")
 (declare-function ghostel--set-size "ghostel-module")
 (declare-function ghostel--write-input "ghostel-module")
+(declare-function ghostel--native-uri-at "ghostel-module")
 
 
 ;;; Automatic download and compilation of native module
@@ -1325,7 +1323,6 @@ when `ghostel-scroll-on-input' is nil.  Call from any path where
 the user's action implies \"show me the prompt\" — typed input,
 paste, yank, drop."
   (when (and ghostel-scroll-on-input ghostel--term)
-    (ghostel--scroll-bottom ghostel--term)
     (setq ghostel--snap-requested t)
     (setq ghostel--force-next-redraw t)))
 
@@ -1570,7 +1567,6 @@ pasted using bracketed paste."
     (ghostel--flush-pending-output)
     ;; CSI H = home, CSI 2 J = erase screen, CSI 3 J = erase scrollback.
     (ghostel--write-input ghostel--term "\e[H\e[2J\e[3J")
-    (ghostel--scroll-bottom ghostel--term)
     (setq ghostel--force-next-redraw t)
     ;; Scrollback is gone; any recorded scroll position no longer
     ;; refers to real content.  Reset so the next redraw anchors
@@ -1851,6 +1847,30 @@ stripped so the copied text matches the original terminal content."
     map)
   "Keymap for clickable hyperlinks in ghostel buffers.")
 
+(defun ghostel--native-link-help-echo (window _ pos)
+  "help-echo handler for OSC8 hyperlinks. Retrieves native URI from libghostty."
+  (with-current-buffer (window-buffer window)
+    (ghostel--native-uri-at-pos pos)))
+
+(defun ghostel--native-uri-at-pos (pos)
+  "Return the native OSC8 hyperlink URI at POS."
+  (save-excursion
+    (goto-char pos)
+    (let* ((line (line-number-at-pos nil t))
+           (total (line-number-at-pos (point-max) t))
+           (row-from-bottom (- total line))
+           (col (current-column)))
+      (ghostel--native-uri-at ghostel--term row-from-bottom col))))
+
+(defun ghostel--uri-at-pos (pos)
+  "Return the URI at POS.
+If the `help-echo' property is a string, return it; otherwise fetch
+the native OSC8 URI at that position."
+  (let ((help-echo (get-text-property pos 'help-echo)))
+    (if (stringp help-echo)
+        help-echo
+      (ghostel--native-uri-at-pos pos))))
+
 (defun ghostel--open-link (url)
   "Open URL, dispatching by scheme.
 file:// URIs open in Emacs; http(s) and other schemes use `browse-url'.
@@ -1879,13 +1899,12 @@ a line suffix opens at the start of the file or directory."
 (defun ghostel-open-link-at-click (event)
   "Open the hyperlink at the mouse click EVENT position."
   (interactive "e")
-  (ghostel--open-link
-   (get-text-property (posn-point (event-start event)) 'help-echo)))
+  (ghostel--open-link (ghostel--uri-at-pos (posn-point (event-start event)))))
 
 (defun ghostel-open-link-at-point ()
   "Open the hyperlink at point."
   (interactive)
-  (ghostel--open-link (get-text-property (point) 'help-echo)))
+  (ghostel--open-link (ghostel--uri-at-pos (point))))
 
 (defun ghostel--find-next-link (from)
   "Return start position of the first hyperlink after FROM, or nil.
@@ -3156,14 +3175,7 @@ position COL columns into the first matched line."
     (when (> tr 0)
       (save-excursion
         (goto-char (point-max))
-        ;; Partial redraws can leave a trailing \n after the last row.
-        ;; Step past it so `forward-line' counts only content rows, not
-        ;; the phantom empty line that would otherwise push the viewport
-        ;; one line too deep and clip the bottom row.
-        (when (and (not (bobp))
-                   (eq (char-before) ?\n))
-          (forward-char -1))
-        (forward-line (- (1- tr)))
+        (forward-line (- tr))
         (line-beginning-position)))))
 
 (defun ghostel--active-preedit-overlay ()
@@ -3310,47 +3322,13 @@ No-op when `ghostel--snap-requested' (user input overrides)."
                    (eq (window-buffer win) buffer))
           (ghostel--reconcile-saved-position win entry))))))
 
-(defun ghostel--anchor-window (win vs pt &optional exact-point)
+(defun ghostel--anchor-window (win vs pt)
   "Pin WIN to viewport-start VS and sync its point to PT.
 Also resets pixel vscroll (pixel-scroll-precision-mode may leave a
-partial offset that would clip the top line after a redraw).
-
-Two terminal-side configurations land PT at `point-max' on the last
-visible row in a position Emacs redisplay treats as off-screen — which
-makes `scroll-conservatively' shift `window-start' up by one row,
-fighting the viewport pin and hiding the block cursor:
-
- 1. Pending-wrap: the last printed character filled the rightmost
-    column and the next print will soft-wrap (issue #138).
-
- 2. CUP park onto an empty trailing row, no pending-wrap: the TUI
-    moved the cursor via absolute positioning to a row that has no
-    written cells, so the row renders to an empty buffer line and PT
-    lands at `point-max' (issue #157).
-
-Clamp `window-point' back by one in either case so it sits inside the
-viewport; buffer-point is unaffected and subsequent redraws recapture
-the real cursor.  We must NOT clamp for a plain shell prompt where the
-cursor is legitimately at `point-max' after typing — doing so would
-draw the block cursor on the last character instead of after it
-\(issue #146).
-
-When EXACT-POINT is non-nil, set `window-point' to PT exactly.  This is
-used while a GUI input method is displaying preedit text: the candidate
-window is anchored to point, and moving it by one character is visible
-as a jump."
+partial offset that would clip the top line after a redraw)."
   (set-window-start win vs t)
   (set-window-vscroll win 0 t)
-  (set-window-point win (if (and (not exact-point)
-                                 (= pt (point-max))
-                                 (> pt (point-min))
-                                 ghostel--term
-                                 (or (ghostel--cursor-pending-wrap-p
-                                      ghostel--term)
-                                     (ghostel--cursor-on-empty-row-p
-                                      ghostel--term)))
-                            (1- pt)
-                          pt)))
+  (set-window-point win pt))
 
 (defun ghostel--restore-scrollback-window (win state)
   "Restore WIN to ws/wp recorded in STATE and push STATE to scroll-positions.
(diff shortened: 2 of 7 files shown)


emil-e:
I will also note that there is much less code now!

dakra:
Oh, wow, thanks.
The zig part is the one that's more or less completely AI. Especially in the beginning of this project.
So I think I'm loving it.
I have a few things in the pipeline but I guess I'll wait now until we land this as so much stuff changes.

You fix the CI first before I have a closer look?

And I see there is a `/// TODO: refactor` in the code.

and I don't think I agree with this statement:

> When libghostty rotates rows out at the scrollback cap, those rows are not evicted from the Emacs buffer. This is intentional: the Emacs buffer serves as the permanent record, and libghostty's scrollback cap only bounds libghostty's memory use, not what is visible in Emacs.

Emacs buffers are generally fast but if you open a 1-200MB file, it does get slow.
And I actually like that you can e.g. run a test or 'cat big-logfile' and everything is as smooth as in ghostty.
Another point is that it could also be confusing. If the Emacs buffer > libghostty scrollback, then you scroll all the way up (M-<) and see something you did in the beginning.. then you just resize a little and suddenly it's gone.
wdyt?

emil-e:
> You fix the CI first before I have a closer look?

I will, I'm iterating as we speak.
 
> And I see there is a `/// TODO: refactor` in the code.

Already gone! 

> and I don't think I agree with this statement:
> 
> > When libghostty rotates rows out at the scrollback cap, those rows are not evicted from the Emacs buffer. This is intentional: the Emacs buffer serves as the permanent record, and libghostty's scrollback cap only bounds libghostty's memory use, not what is visible in Emacs.
> 
> Emacs buffers are generally fast but if you open a 1-200MB file, it does get slow. And I actually like that you can e.g. run a test or 'cat big-logfile' and everything is as smooth as in ghostty. Another point is that it could also be confusing. If the Emacs buffer > libghostty scrollback, then you scroll all the way up (M-<) and see something you did in the beginning.. then you just resize a little and suddenly it's gone. wdyt?

I don't have a strong opinion on this so I'm happy to just mirror whatever libghostty is doing. It has the added benefit that OCS8 links far back in scrollback don't lose their URI once libghostty evicts them. I agree with your suggestion, it's a better behavior.

emil-e:
Need a bit to implement the scrollback eviction.

dakra:
Awesome. I'll test drive this a bit.

2 more quick things:
- There are tabs in ghostel(-test).el -> M-x untabify
- `make checkdoc` complains (I should fix the CI so it errors there as well)

emil-e:
@dakra Checkdoc had a lot of complaints unrelated to my chnage so I didn't want to just sprinkle random fixes in this one. Please tell me if you still want me to change them.

emil-e:
I also see that one test is failing, but that seems to just be flakiness. Works On My Machine (TM)

dakra:
Awesome. I'm very excited for this.

Locally `make test-evil` fails. The CI is somehow green. does it work for you?
I just spotted it by `make -j4 all` that I always do to check if everything is working.

emil-e:
> Awesome. I'm very excited for this.

I'm glad to hear. It's a shame that performance is going down a bit, but I think correctness is more important and we can optimize further. I ran Apple Instruments a bunch while building this and I think that there might be some tricks we can use to gain some of it back, as I said.

> Locally `make test-evil` fails. The CI is somehow green. does it work for you? I just spotted it by `make -j4 all` that I always do to check if everything is working.

`make -j4 test-evil` fails for me too. It probably has to do with the trailing newline. I can take a look.

emil-e:
It was indeed the trailing newline. Adjusted the test.

dakra:
@emil-e I only tried it a little bit but didn't find anything wrong so far. code looks good as well and I'm always a fan of simplifications.
Should I rebase and land this on the main branch? What do you think?

emil-e:
@dakra I also haven't found anything wrong with it when I used it today at work, but I have found things wrong with main. And the tests pass. I say ship it, personally.

dakra:
ok.
I'll ship it in an hour when I'm home.

claude review found something if you want to have a look:
```
  - src/module.zig:1238 fnUriAt underflow — getTotalRows() - row_from_bottom underflows when a hovered link sits in
  materialized scrollback past libghostty's cap; also unchecked @intCast to usize/u16. Add bounds check + clamp.
  - src/render.zig:741 spurious full-redraw at 1-row scrollback — scrollbar_hit_cap is true on every redraw when scrollback is
   exactly 1 row; comment block at 730-738 hints the intended guard total > len + 1 but it's missing.
```

and gpt review found:
```
2. Medium: extensions/evil-ghostel/evil-ghostel.el:88-91
     evil-ghostel--reset-cursor-point now overcounts scrollback by one line.

     Why it looks wrong:
     The renderer now always materializes a trailing newline for the viewport. count-lines (point-min) (point-max) includes that
     phantom final line, so:
     count-lines - ghostel--term-rows
     reports one line of scrollback even when there is none.

     User-visible effect:
     In a full viewport with no actual history, Evil point restoration can jump one row too far down when syncing point to the
     terminal cursor.

     What I would change:
     Stop deriving scrollback from raw buffer line count. Use the actual viewport anchor instead, e.g. compute scrollback from
     ghostel--viewport-start, which already handles the trailing-newline shape correctly.
```

emil-e:
@dakra They look like they could real issues, at least at a glance. I'll have a look a bit later.

emil-e:
The Evil one is hard for me to check since I don't even know how Evil mode works. The other ones are real issues that I will take care of. Perhaps hold off on merging until I fix them.

emil-e:
@dakra Added the bounds checks. The scroll edge case needs a bit more thinking but I think it's fine to merge as is since it's at worst an extra redraw, but it's a redraw at a small size when there is only one scrollback line to it's likely fast anyway. If I don't come up with something before you merge I'll make a followup PR.

The Evil one I'm not sure I'm qualified to fix.

emil-e:
Here's what Sonnet said about the evil thing:
```
⏺ The GPT review is wrong about count-lines. Let me explain why:             
                                                                          
  How count-lines handles trailing newlines                                  
                                                                             
  From the Emacs docs: count-lines returns the number of newlines between    
  START and END, and adds one only if the text does not end with a newline.  
  Since the buffer now always has a trailing \n, count-lines returns exactly 
  the number of newlines — no phantom extra line.                            
                                                                             
  Example for a full 3-row viewport with no scrollback:                      
  line1\nline2\nline3\n
  - count-lines = 3 (three newlines, buffer ends with \n → no +1)            
  - scrollback = max(0, 3 - 3) = 0 ✓                             
                                                                             
  The old bug was line-number-at-pos (point-max): when the buffer ends with
  \n, point-max sits at the start of an empty line N+1, so line-number-at-pos
   returned N+1 — that was the actual off-by-one. Commit 6839511 fixed it
  correctly.                                                                 
                            
  The GPT's proposed fix (using ghostel--viewport-start) would also work, but
   it's unnecessary complexity — count-lines already has the right semantics
  here.                                                                      
                            
  The current code at lines 88-89 is correct.   
  ```

**Comment by emil-e** [comment]
Regarding this:
```
  - src/render.zig:741 spurious full-redraw at 1-row scrollback — scrollbar_hit_cap is true on every redraw when scrollback is
   exactly 1 row; comment block at 730-738 hints the intended guard total > len + 1 but it's missing.
```

This is a real thing, but it's not trivial to fix. There are some things that conflict here:
- We position the viewport at `max_offset - 1` to get two things - a signal for scrolling and a signal for scrollback reset.
- We clear everything if the viewport ends up at `offset == 0` because then we might have scrolled past the end of the scrollback if we received a very large VT write.

Since we don't know exactly where the cap is, there's no surefire way for us to know whether `offset == 0` is just normal scrolling or we're at the cap.

The only way I can think of right now is to not use `offset == 0` as the signal for capping but instead always keep a an additional render state that we render from the top. When the top row of that render state gets dirty, we know that the scrollback is capped and invalid because that is the only way that scrollback can change. But I'm not sure it's better than invalidating a few extra lines in case we receive a bunch of data when we have exactly one row of scrollback.

**Comment by dakra** [comment]
@emil-e I run a claude /ultrareview and it flagged 2 things:

----

Review complete — 2 findings

🔴 src/module.zig:1038-1046

The viewport-restore defer in fnCursorPosition (and fnDebugState/fnDebugFeed) sits inside an if (term.getScrollbar()) |sb| { defer { ... } } block whose body is just the defer registration, so the defer fires the instant the block ends — before the SCROLL_BOTTOM call below it. Net effect: the function leaves the viewport at the bottom (offset + len == total), and the next redraw() mistakes that for a scrollback-clear signal and triggers a full erase + rebuild, which is particularly bad under evil-ghostel where ghostel--cursor-position runs on every state transition. Fix by capturing sb.offset into a function-scoped optional and putting the restore defer at function scope.

🔴 src/module.zig:1255-1269

Use-after-free in fnUriAt heap fallback path. The defer std.heap.c_allocator.free(buf) on line 1259 is lexically scoped to the inner if (alloc) |buf| block, so buf is freed when that block exits at line 1264 — but heap_uri still aliases the freed slice, and env.makeString(uri[0..out_len]) on line 1268 then reads through it. Hoist the defer to function scope, e.g. defer if (heap_uri) |buf| std.heap.c_allocator.free(buf); — the same pattern the removed scanHyperlinksFromGrid used. While here, also check the second ghostty_grid_ref_hyperlink_uri returned SUCCESS before using its output.

---

then I tasked claude to check and create a fix:
```
Both findings are real. Here's my read and a plan to fix.

  Verification

  Finding 1 — viewport-restore defer fires too early (module.zig:1038-1046, plus identical bug in fnDebugState:903-910 and fnDebugFeed:969-976)

  Confirmed. In Zig, defer runs at the end of the block that registered it, not the enclosing function. The if (term.getScrollbar()) |sb| { defer { ... } } block has nothing in it but the defer registration, so the defer fires the
  moment the if block exits — before scrollViewport(SCROLL_BOTTOM, 0) on the next line. Net effect: each call leaves the viewport at the bottom.

  The blast-radius claim also checks out — render.zig:740 explicitly treats scrollbar.len + scrollbar.offset == scrollbar.total as the "libghostty cleared scrollback, full erase + rebuild" signal, and evil-ghostel.el:86/97/181 calls
  ghostel--cursor-position on every state transition. So with scrollback present, every i/Esc causes a buffer wipe and full redraw.

  Finding 2 — use-after-free in fnUriAt heap fallback (module.zig:1255-1269)

  Confirmed. defer std.heap.c_allocator.free(buf) on line 1259 is scoped to the if (alloc) |buf| block and fires when that block exits at line 1264. heap_uri still aliases the freed slice, and env.makeString(uri[0..out_len]) on line
  1268 reads through it. Pure UB — works only because the allocator hasn't reused the chunk yet.

  The reviewer's note about checking the second ghostty_grid_ref_hyperlink_uri for SUCCESS is also valid: today an error from the second call would leave result != SUCCESS but we still call makeString(uri[0..out_len]) with whatever
  out_len ended up as. The first call has the same gap (e.g., result == NOT_FOUND should return nil, not stringify garbage).

  Line 126 in the same file already shows the correct pattern: defer if (heap_buf) |hb| std.heap.c_allocator.free(hb); at function scope.

  Fix plan

  1. module.zig:1034 fnCursorPosition — replace the inner-block defer with a function-scoped optional:
  const saved_offset: ?usize = if (term.getScrollbar()) |sb| sb.offset else null;
  defer if (saved_offset) |off| {
      term.scrollViewport(gt.SCROLL_TOP, 0);
      term.scrollViewport(gt.SCROLL_DELTA, @intCast(off));
  };
  term.scrollViewport(gt.SCROLL_BOTTOM, 0);
  1. → verify: with scrollback parked above bottom, two consecutive cursor-position calls do not leave the viewport at offset + len == total.
  2. module.zig:899 fnDebugState — same edit.
  3. module.zig:965 fnDebugFeed — same edit.
  4. module.zig:1229 fnUriAt — hoist the free to function scope and add SUCCESS checks on both calls:
  var heap_uri: ?[]u8 = null;
  defer if (heap_uri) |buf| std.heap.c_allocator.free(buf);
  ...
  if (result == gt.OUT_OF_SPACE and out_len > uri_stack.len) {
      const buf = std.heap.c_allocator.alloc(u8, out_len) catch return env.nil();
      heap_uri = buf;
      result = gt.c.ghostty_grid_ref_hyperlink_uri(&grid_ref, buf.ptr, buf.len, &out_len);
  }
  if (result != gt.SUCCESS or out_len == 0) return env.nil();
  const uri = if (heap_uri) |buf| buf else &uri_stack;
  return env.makeString(uri[0..out_len]);
  4. → verify: existing make test-native URI tests still pass; manually hover a >2KB URI (heap path) to exercise the fallback.
  5. Regression test for finding 1. Easiest is a native test that:
    - feeds enough rows to create scrollback,
    - parks the viewport above bottom via ghostel--scroll-to,
    - calls ghostel--cursor-position,
    - asserts the scrollbar offset is still where we parked it (i.e., offset + len < total).

  Without this, we have no guard against future regressions of the same shape.
  ```
  
  I let claude implement it and commit it on here.
Do you want to check? You can revert it if it's rubbish.

If you give the thumbs up, I'm squashing the commits and put it on main.

emil-e:
@dakra Fixes look good to me, but I would remove comments about why it's at function scope - even though I apparently don't know Zig that well (C++ and Rust is my home turf) one should probably assume that whoever reads the code know `defer` works. Agents have a tendency to leave these "oddly specific" comments around making it pretty confusing after a while.

dakra:
> @dakra Fixes look good to me, but I would remove comments about why it's at function scope - even though I apparently don't know Zig that well (C++ and Rust is my home turf) one should probably assume that whoever reads the code know defer works. Agents have a tendency to leave these "oddly specific" comments around making it pretty confusing after a while.

Agree. Especially Claude is very verbose with it's comments. I removed them again.

dakra:
Awesome PR. Thank you very much.
Let's see how it works for other people!

~~Next I would like to remove incremental redraw path (`ghostel-full-redraw`).
That was still from the very beginning where we didn't materialize to an Emacs buffer yet.
I think incremental redraw is now really not worth it to keep around.
Do you have an opinion on that?
I would make a PR and we discuss there?!~~

Scratch that. The existing tui benchmark always updates ALL rows, that's why it's just the same
speed as a full redraw.
But I just added a benchmark for something like just a statusbar update (1 row change)
and there the incremental redraw is 10-15x faster.

emil-e:
@dakra Yes, according to my profiling, the bottleneck is `buildRowContent`, especially reading styles and comparing them for style runs. I have an idea there to speed it up a lot that I’m going to try.

I still think the incremental render is worth it for now since it’s not a lot of code - if it’s not dirty we `forward-line` and now it seems very reliable to me.

PR #206, opened 2026-04-29
Use 80x24 default size when ghostel-exec target buffer is undisplayed

dakra:
## Summary

- `ghostel-exec` previously fell back to `(selected-window)` when BUFFER was not displayed in any window — sizing the PTY from whatever window happened to be focused at call time, which rarely matches where BUFFER eventually shows up.
- Match `eat`'s behavior: when no window displays BUFFER, start at the universal 80x24 default. The displayed-buffer path is unchanged; SIGWINCH still resizes the PTY when the displaying window's dimensions differ from the initial size.
- Reported by a user running a programmatic flow that creates a buffer with `generate-new-buffer` and immediately calls `ghostel-exec` before displaying it — the agent (Pi) consistently got the wrong terminal dimensions.

## Test plan

- [x] `make -j4 all` passes (39 zig + 100 native + 193 elisp tests, lint clean).
- [x] New ERT test `ghostel-test-exec-uses-default-size-when-buffer-not-displayed` covers the fix and would fail under the old selected-window fallback.
- [x] Manual verification: in a live emacsclient session, invoke `ghostel-exec` on a freshly-created undisplayed buffer (e.g., `(let ((b (generate-new-buffer "x"))) (ghostel-exec b "less" '("/etc/hosts")) b)`) and confirm the buffer-local `ghostel--term-rows` / `ghostel--term-cols` are 24 / 80 before display, and that the program lays out correctly once the buffer is shown in a larger window.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -3687,12 +3687,12 @@ Returns the buffer."
   "Run PROGRAM with ARGS as a ghostel terminal in BUFFER.
 
 BUFFER is switched into `ghostel-mode' (if not already) and a new
-terminal is created sized to the window displaying BUFFER, falling
-back to the selected window if BUFFER is not displayed.  No shell
-integration is applied — PROGRAM is exec'd directly via
-`ghostel--spawn-pty'.  PROGRAM is shell-quoted before it is passed
-to `/bin/sh -c', so shell metacharacters are not interpreted; pass
-extra tokens via ARGS, a list of strings.  Returns the process.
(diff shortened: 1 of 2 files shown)


emil-e:
I'm not sure this will solve my problem at least, since this might just be a different incorrect size. But it's still better than using selected window so I think it's a good change.

dakra:
OK, I merged it.
Let's open a new issue if that shouldn't fix your problem.

PR #207, opened 2026-04-29
Commit deferred minibuffer crop on output (fix #191)

### LINKED ISSUE #191: M-x with vertico while scrolling can cause "cursor following" to stop
## Repro steps
1. Start continuously scrolling process - such as an autonomous agent session in Pi coding agent.
2. Make It full height in the Emacs frame.
3. Activate M-x with something (like vertico) that expands the minifbuffer. This resizes the Ghostel window.
4. Cancel C-g

## Expected
Terminal keeps scrolling along
## Actual
Terminal stops scrolling at the point where the minibuffer closed

## Comments
The scrolling keeps working when the minibuffer opens, it's when it closes that it breaks. This is not always consistent, but pretty easy to repro, at least in Pi.

dakra:
## Summary

- Fixes #191: cursor following stops following the viewport after a minibuffer cycle (e.g. `M-x` with vertico) on a continuously-scrolling terminal.
- Root cause: the CROP branch of `ghostel--window-adjust-process-window-size` returned nil to suppress SIGWINCH but left `ghostel--term-rows` at the pre-crop size. When the minibuffer closed and the window grew back, the next adjust callback saw `eql height ghostel--term-rows` and hit the no-change branch, skipping the resize-aware redraw — `keep-point-visible` drift on `window-start` could then be misclassified as user scrollback, latching the auto-follow predicate off.
- Fix: the CROP branch now stores the pending size in a new buffer-local `ghostel--deferred-crop-size`. `ghostel--commit-deferred-crop` flushes it before the next render, called from `ghostel--filter` on the next chunk of output. The optimization (no SIGWINCH for a quiet minibuffer flash) is preserved — the deferred size is cleared without committing if the window grows back before any output arrives. The no-change / real-resize branches and `ghostel--commit-cropped-size` clear the flag when the deferred size is moot or superseded.

## Test plan

- [x] `make -j4 all` (zig build + 219 elisp + 103 native + 39 zig tests, lint, byte-compile)
- [x] `make test-evil` (39 evil-ghostel tests)
- [x] New test `ghostel-test-191-deferred-crop-commit-prevents-anchor-latch` — sets the deferred size, simulates `window-start` drift below the anchor, asserts post-commit: `term-rows` updated, flag cleared, exactly one SIGWINCH, `scroll-positions` empty, `window-start` pinned at the new viewport-start.
- [x] New test `ghostel-test-191-no-output-clears-deferred-crop-on-regrow` — exercises the no-output regrow path, asserts the no-change branch clears the flag without firing SIGWINCH.
- [x] Live verification: full-height frame, continuous-output process, `M-x` with vertico, `C-g` — confirm scrolling continues post-cancel and `ghostel--scroll-positions` stays empty.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -920,6 +920,25 @@ Updated whenever the terminal is created or resized.")
   "Column count of the native terminal.
 Updated whenever the terminal is created or resized.")
 
+(defvar-local ghostel--deferred-crop-size nil
+  "Pending (HEIGHT . WIDTH) from a deferred minibuffer-shrink crop.
+Set by the CROP branch of `ghostel--window-adjust-process-window-size'
+when the minibuffer shrinks the ghostel window — we suppress SIGWINCH
+for transient minibuffer flashes with a quiet terminal but record the
+size so it can be committed if anything would render at the wrong
+geometry.  Committed by `ghostel--commit-deferred-crop' when output
+arrives, and cleared without committing when the window grows back to
+or beyond `ghostel--term-rows' before output arrives.
+
+Without this, `ghostel--term-rows' stays at the original (pre-crop)
+size for the duration of the minibuffer.  When the minibuffer closes
+and the window grows back, `ghostel--window-adjust-process-window-size'
+sees `eql height ghostel--term-rows' and hits the \"no change\" branch,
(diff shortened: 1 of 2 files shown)


dakra:
Superseeded by #211

PR #208, opened 2026-04-29
Detect style run breaks via cheap CellStyleKey rather than full CellStyle comparison

emil-e:
## What

The hot inner loop in `buildRowContent` previously called `readCellStyle` on
every terminal cell and compared the result field-by-field with `CellStyle.eql()`
across all ten style attributes (fg, bg, bold, italic, faint, underline,
underline_color, strikethrough, inverse, hyperlink).

This PR replaces that with a two-field `CellStyleKey` (`style_id` + `hyperlink`)
that is cheap to read and compare. The full `readCellStyle` call is now deferred
to run boundaries only — once per style change rather than once per cell.

`style_id` is Ghostty's unique identifier for a cell's complete attribute set.
`hyperlink` is tracked separately because Ghostty does not fold hyperlink state
into the style ID, but Ghostel uses text properties for it.

Supporting changes:

- `RunInfo.style` becomes `?CellStyle` (null = default style), removing the
  `isDefault()` guard from `applyStyle` and making the null check explicit at
  the call site
- Wide-character spacer tails compensate with `end_char -= 1` after the style
  logic runs, preventing boundary overcounting that would otherwise bleed style
  from a styled wide character onto the following default-style cell
- `CellStyle.eql()` is removed; `std.meta.eql` on `CellStyleKey` replaces it
- The explicit "close final run" block at the end of the loop is no longer needed

## Benchmark

All runs: `zig build -Doptimize=ReleaseFast`, Elisp byte-compiled.  
Machine: Apple M5 Pro, macOS 26.4.1, Emacs 30.2.

### Streaming (chunked write + periodic redraw, no PTY)

| Scenario    | main   | branch | Δ        |
|-------------|-------:|-------:|---------:|
| stream/incr | 25 ms  | 20 ms  | **−20%** |
| stream/full | 26 ms  | 20 ms  | **−23%** |

### TUI frame rendering (full-screen rewrites)

| Scenario    | main       | branch     | Δ        |
|-------------|-----------:|-----------:|---------:|
| incr 24×80  | 11 379 fps | 13 466 fps | **+18%** |
| full 24×80  | 11 448 fps | 13 805 fps | **+21%** |
| incr 40×120 |  5 182 fps |  6 550 fps | **+26%** |
| full 40×120 |  5 285 fps |  6 539 fps | **+24%** |

### TUI partial update (bottom-row update over static screen)

| Scenario    | main        | branch      | Δ        |
|-------------|------------:|------------:|---------:|
| full 24×80  |  14 180 fps |  18 110 fps | **+28%** |
| full 40×120 |   6 801 fps |   9 634 fps | **+42%** |
| incr 24×80  | 124 031 fps | 133 969 fps |    +8%   |
| incr 40×120 |  87 782 fps |  99 775 fps |   +14%   |

### Engine micro-benchmarks (single bulk call, no PTY)

| Scenario     | main     | branch   | Δ     |
|--------------|----------:|---------:|------:|
| plain/incr   | 128 MB/s | 135 MB/s |  +5%  |
| styled/incr  |  89 MB/s |  95 MB/s |  +7%  |
| unicode/incr | 179 MB/s | 197 MB/s | +10%  |

### Real-world PTY (cat 1 MB through process pipe)

| Scenario   | main     | branch   | Δ    |
|------------|----------:|---------:|-----:|
| plain/incr | 105 MB/s | 111 MB/s | +6%  |
| plain/full | 107 MB/s | 110 MB/s | +3%  |

PTY improvements are modest: throughput there is dominated by process I/O and
timer overhead rather than per-cell render cost.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -5,12 +5,17 @@ const vendored_emacs_module_dir = "vendor";
 pub fn build(b: *std.Build) void {
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
+    const ghostty_optimize = b.option(
+        std.builtin.OptimizeMode,
+        "ghostty-optimize",
+        "Optimization mode for the ghostty dependency (defaults to the main optimize option)",
+    ) orelse optimize;
     const is_release = optimize != .Debug;
     const target_os = target.result.os.tag;
     const emacs_module_dir = resolveEmacsModuleDir(b);
     const ghostty_dep = b.dependency("ghostty", .{
         .target = target,
-        .optimize = optimize,
+        .optimize = ghostty_optimize,
         .@"emit-lib-vt" = true,
     });
 
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,6 +1,6 @@
 .{
     .name = .ghostel,
-    .version = "0.20.0",
+    .version = "0.20.1",
     .paths = .{""},
     .fingerprint = 0x5a44bdd1198a0f4b,
     .dependencies = .{
--- a/extensions/evil-ghostel/evil-ghostel.el
+++ b/extensions/evil-ghostel/evil-ghostel.el
@@ -4,7 +4,7 @@
 
 ;; Author: Daniel Kraus <daniel@kraus.my>
 ;; URL: https://github.com/dakra/ghostel
-;; Version: 0.20.0
+;; Version: 0.20.1
 ;; Package-Requires: ((emacs "28.1") (evil "1.0") (ghostel "0.8.0"))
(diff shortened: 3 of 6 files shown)


emil-e:
Flaky test, as far as I can tell.

dakra:
Very nice. Thanks.

BTW, if you want I can give you maintainer rights here on this repo.
Then you could trigger the flaky pipeline yourself and also close PRs/Issues etc.
But don't feel obligated to help maintain, just in case you're interested since you're also
very invested in ghostel already :)

emil-e:
@dakra I am indeed pretty invested by now and I'd love to have a chat about. I'll drop you an e-mail!

PR #211, opened 2026-04-30
Only rebuild scrollback on column size changes

emil-e:
## Summary

- **Only rebuild on column size changes** — vertical-only resizes (row count up or down) no longer erase and rebuild the scrollback buffer. Only column changes (which invalidate line wrapping) trigger a full rebuild. Resizes are now deferred via `pending_resize` and committed at redraw time so the redrawer sees the full pre-resize active area before committing.
- **Remove minibuffer hacks** — the `ghostel--commit-cropped-size` / crop-deferral logic that tried to hide resize events while the minibuffer was open is no longer needed now that resize commits are deferred; removed.
- **Tests** — 7 new ERT tests covering scrollback survival across shrinks, expands, roundtrip resize sequences, and incremental dirty-row reuse.

## Key changes

- `terminal.zig`: `cols`/`rows` → `size: ViewportSize`; new `pending_resize` field; `cell_width_px`/`cell_height_px` stored on terminal; `resize()` no longer calls `ghostty_terminal_resize` directly.
- `render.zig`: `commitResize()` called at redraw time; rebuild condition is `cols_changed` (not `rebuild_pending`); `rows_in_buffer` replaces `scrollback_in_buffer` (tracks total rows in buffer, not just scrollback rows); `renderToEnd()` extracted.
- `ghostel.el`: `ghostel--commit-cropped-size` and minibuffer crop path removed; `ghostel--set-size-with-cell-dims` wrapper keeps resize call sites consistent.

## Test plan

- [x] `make test` passes (ERT suite including new tests)
- [x] Resize a ghostel buffer vertically — scrollback content is preserved without flicker
- [x] Resize a ghostel buffer horizontally — scrollback is rebuilt (expected)
- [x] Open minibuffer while ghostel is visible — no resize artifact on close

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -4023,18 +4023,6 @@ PROCESS is the shell process, WINDOWS is the list of windows."
            ((and (eql height ghostel--term-rows)
                  (eql width ghostel--term-cols))
             (setq size nil))
-           ;; Crop: minibuffer is active, only height shrank, width unchanged, and the
-           ;; terminal is on the primary screen.  Defer the resize so no SIGWINCH is sent
-           ;; — the Emacs window naturally hides the bottom rows, just like a normal
-           ;; buffer.  Alt-screen apps (vim, htop) own the full viewport and must be told
-           ;; the real size to re-layout, so they take the normal-resize path.  If the
-           ;; user switches focus into the ghostel window while the minibuffer is still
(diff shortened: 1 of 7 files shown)


dakra:
Awesome 🚀 

BTW, replied to your mail yesterday. no rush, but just in case it landed in spam (as gmail sometimes does to external providers :(

PR #216, opened 2026-05-01
Route ESC to terminal in alt-screen apps under evil-ghostel (#215)

### LINKED ISSUE #215: Sending ESC to ghostel with evil-mode on
e.g. run ghostel with evil-mode -> start up vim , press ESC

Since ESC is special in evil-mode, emacs will not send ESC to the terminal

We do this for vterm https://github.com/emacs-evil/evil-collection/blob/8db8a959d2c1ca54496224e96ff712c9185e38ab/modes/vterm/evil-collection-vterm.el#L58 as a quick hack.

dakra:
## Summary

Fixes #215. Pressing ESC in evil insert state inside a ghostel buffer ran `evil-normal-state` even when the inner app (vim, less, htop, nvim, etc.) needed the keystroke.

- New buffer-local routing mode controlled by `evil-ghostel-escape`: `auto` (default) | `terminal` | `evil`. `auto` checks DECSET 1049 (alt-screen) to decide.
- Interactive `evil-ghostel-toggle-send-escape`: no-arg cycles `auto → terminal → evil → auto`; numeric prefix `1`/`2`/`3` sets directly. Out-of-range numeric prefix signals `user-error`.
- Terminal-bound ESC snaps the viewport via `ghostel--snap-to-input` to match the invariant every other key in `ghostel-mode-map` holds.
- Evil-bound fallback lands on `evil-force-normal-state` when the user's `<escape>` binding is missing or a chord prefix (e.g. `evil-escape`'s `jk`), so the keystroke is never silently dropped.

## Test plan

- [x] `make -j4 all` — 218 ERT tests pass, 11 new in `test/evil-ghostel-test.el`:
  - dispatcher branches: terminal-always / evil-always / auto+altscreen / auto+no-altscreen
  - viewport snap on terminal-bound ESC
  - evil fallback when `lookup-key` returns nil (chord/rebound case)
  - toggle cycle, numeric prefix set, invalid prefix `user-error`
  - per-buffer isolation (sets `terminal` in buffer A, `evil` in buffer B, re-enters A)
  - mode init reads `evil-ghostel-escape` defcustom
- [x] Live verify in `M-x ghostel`:
  - `vim` insert mode → ESC leaves insert (auto path).
  - Toggle to `evil` mid-vim, ESC switches evil state instead.
  - `opencode` (alt-screen but user wants evil ESC) → toggle to `evil` works.
  - Shell prompt with no alt-screen → ESC switches evil to normal (auto path, default).
  - `M-2 evil-ghostel-toggle-send-escape` directly forces `terminal`.

### DIFF
--- a/extensions/evil-ghostel/evil-ghostel.el
+++ b/extensions/evil-ghostel/evil-ghostel.el
@@ -56,6 +56,22 @@ last-writer-wins."
          (set-default-toplevel-value sym val)
          (evil-set-initial-state 'ghostel-mode val)))
 
+(defcustom evil-ghostel-escape 'auto
+  "Where insert-state ESC is routed in ghostel buffers.
+
+`auto'      — when the inner app is in alt-screen mode (DECSET 1049,
+              used by vim, less, htop, nvim, etc.) ESC is sent to the
+              terminal; otherwise evil's binding runs and switches to
+              normal state.
+`terminal'  — always send ESC to the terminal.
+`evil'      — always run evil's binding (ESC stays with evil).
+
+Sets the initial value of the buffer-local state.  Use
+\\[evil-ghostel-toggle-send-escape] to change it for the current buffer."
+  :type '(choice (const :tag "Auto (alt-screen heuristic)" auto)
+                 (const :tag "Always to terminal" terminal)
+                 (const :tag "Always to evil" evil)))
+
 ;; Apply the current value at load.  Covers the case where the user set
 ;; the variable with plain `setq' before loading the package — in that
 ;; path `defcustom' preserves the value without invoking `:set'.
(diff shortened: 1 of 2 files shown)


extensions/evil-ghostel/evil-ghostel.el
```diff
@@ -368,6 +384,62 @@ ORIG-FN is the advised `evil-redo' called with COUNT."
...
+                   (user-error
+                    "Invalid prefix %d; use 1 (auto), 2 (terminal), or 3 (evil)"
+                    n)))
+           (let ((next (cdr (memq evil-ghostel--escape-mode
+                                  evil-ghostel--escape-modes))))
+             (or (car next) (car evil-ghostel--escape-modes))))))
+    (setq evil-ghostel--escape-mode target)
+    (message "evil-ghostel ESC mode: %s" target)))
+
+(evil-define-key* 'insert evil-ghostel-mode-map
```
jojojames, line 440: I'd imagine there's a scenario where you're in normal state in emacs but want ESC to be sent to the insert state in the nested emacs/vim or maybe vice versa. Not sure because we're dealing with 2 modal states that could be not synchronized.

Or maybe in a non normal/insert state (like visual/motion/etc?) state but wanting ESC to pass through.

dakra: Since I'm not an evil user I'm not 100% I understand what you mean, can you rephrase or make a suggestion for a fix?

btw (maybe it helps), you can always `C-c C-q <esc>` to send the next key to the terminal whatever it is (<esc> in this case).

jojojames: e.g. if you're in emacs (outer) and vim (inner) [start emacs -> start ghostel -> start vim inside ghostel

maybe using 'insert state is ok e.g. for the user to interact with ghostel, they need to be in insert state, which then triggers your function above

the cycling might be strange though

e.g. emacs+evil [normal state] -> go to insert state -> press i in vim to go to insert state -> press ESC. this cycles the ESC to the next thing.

But you want this instead:

emacs+evil [normal state] -> go to insert state in emacs -> press i in vim to go to insert state -> press ESC -> (normal state in vim) -> press i again to go to insert state in vim -> press ESC to go to normal state in vim -> press i again to go to insert state in vim -> and so on.

^ I think this case is more likely to happen.

If you do that though, you still need a way to get back to the outer emacs (e.g. to get back to normal state in emacs) but once the inner vim is taking ESC inputs, it really won't know which ESC you want to send to so you can't do it automagically.

> suggestion for a fix?

No suggestion, the way we do it in vterm is kind of a "this is too complicated, just let the user decide when they want what". If there's a more clever way to handle this, I'm all for it.

dakra: > emacs+evil [normal state] -> go to insert state in emacs -> press i in vim to go to insert state -> press ESC -> (normal state in vim) -> press i again to go to insert state in vim -> press ESC to go to normal state in vim -> press i again to go to insert state in vim -> and so on.

> ^ I think this case is more likely to happen.

> If you do that though, you still need a way to get back to the outer emacs (e.g. to get back to normal state in emacs) but once the inner vim is taking ESC inputs, it really won't know which ESC you want to send to so you can't do it automagically.

But doesn't the auto-mode kinda address that in that is sends the ESC so long to vim as long as vim is open. You close it to type something in the shell, ESC gets routed to evil?!

And if you want your first case or not have ESC forwardet to the terminal when an alt-screen app is running, you have to manually change the routing (like in your vterm solution right now)?!

jojojames: You might want to leave the window/leave vim running but still be able to get back to Normal state in emacs. I think this is a relatively common use case so won't describe specific examples. :)

> And if you want your first case or not have ESC forwardet to the terminal when an alt-screen app is running, you have to manually change the routing (like in your vterm solution right now)?!

Yeah, I think it's a more manual/intentional step.

dakra: > You might want to leave the window/leave vim running but still be able to get back to Normal state in emacs. I think this is a relatively common use case so won't describe specific examples. :)

ok.
But anyway, should we merge this PR as it provides at least a workaround (and with the auto-mode hopefully even a slightly better hack than vterm)?
(and this might be me not using it, but if you just want to get into normal state emacs, you
can of course just bind `evil-force-normal-state` so something like `C-c esc` or whatever.)

And if you or some other evil user has a better idea how that "workflow" could be implemented I'm open for issues or PRs.

jojojames: Yeah, sounds good.

dakra: Great, thanks.

And happy to discuss any ideas/PRs, especially evil related as I don't test this integration very much.

PR #220, opened 2026-05-02
Refactor Zig code to use array type instead of bare arrays all over the place

emil-e:
Safer code is better code. Chose not to use Zig builtin types because the ergonomics are poor and require the buffer to be separate.

### DIFF
--- a/src/fixed_array_list.zig
+++ b/src/fixed_array_list.zig
@@ -1,0 +1,179 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+/// Minimalist array list implementation with fixed inline storage. Exists due
+/// to poor ergonomics with builtin Zig stdlib types and not wanting external
+/// dependencies.
+pub fn FixedArrayList(comptime T: type, comptime buffer_capacity: usize) type {
+    return struct {
+        const Self = @This();
+
+        pub const capacity = buffer_capacity;
+
+        buffer: [capacity]T = undefined,
+        len: usize = 0,
+
+        /// Appends a single item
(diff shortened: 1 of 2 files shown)


emil-e:
@dakra Any input or should I just merge?

dakra:
For me the code looks great. Some claude review comments, do with them what you like:

- `ghostel--has-wide-chars` is now set per-row instead of once per render. Old code aggregated has_wide_chars across rows and called `env.call2(intern("set"), …, t)` exactly once after the loop (gated on `dirty != DIRTY_FALSE)`. New code calls it inside `insertAndStyle` for every dirty row that contains a wide cell. Functionally equivalent (idempotent set), but it adds an `intern + call2` per such row. Minor, but trivially recoverable: keep the old return-the-flag pattern, or stash the

**fixed_array_list.zig**

- Consider std.BoundedArray(T, N) from the stdlib — it's the same idea (inline buffer + len, append/slice/constSlice/resize, error.Overflow). The PR description says builtin types have "poor ergonomics," but BoundedArray is exactly the shape you built. If there's a concrete reason it doesn't fit (missing method, Zig 0.15.2 API gap), worth a one-line comment in the file; otherwise dropping the helper would save ~180 LOC.
- unusedCapacitySlice simplifies to self.buffer[self.len..capacity]. The self.len + self.unusedCapacity() form is correct
  but indirect.
- Typo: // Returns a slice of the reamining capacity. → remaining (src/fixed_array_list.zig:39).
- items() and constItems() have identical doc comments; one says "mutable" the other "read-only" would help.
- Missing blank line between the last two tests (constItems and resize: shrinks…), src/fixed_array_list.zig:171.

PR #225, opened 2026-05-03
Treat everything as property runs, including prompt and input

emil-e:
This simplifies the code a great deal and makes it easier to add further aspects to a property run.


emil-e:
@dakra Want to have a look or should I just merge?

dakra:
just merge

PR #234, opened 2026-05-04
Match ghostty's OSC 133 marker layout for bash/zsh/fish

### LINKED ISSUE #230: Cursor/typing is on line after prompt
Not sure on repro yet, but it seems it was either introduced today in ghostel or in the last 100 or so commits in Emacs

Image

dakra:
## Summary

- Closes #230. Cursor was landing on a blank row below the prompt char under powerlevel10k's multi-line prompt because the `zle-line-init` fallback emitted raw 133;A from a context where the cursor had already moved past column 0 — and libghostty CR+LFs on 133;A in that state ("fresh-line" behavior). Ghostty's own integration uses 133;P (no fresh-line) for that exact reason.
- Realigns all three shells to match ghostty's marker emission. Most other ghostel-specific code (OSC 51 elisp eval, SSH terminfo wrapper) is untouched.

## Per-shell changes

**zsh** (`etc/shell/ghostel.zsh`)
- `mark1=$'%{\\e]133;A;cl=line\\a%}'`, `mark2=$'%{\\e]133;P;k=s\\a%}'` (multi-line + PS2), `markB=$'%{\\e]133;B\\a%}'` — all BEL-terminated.
- Wrap PROMPT2; inject mark2 after each newline in PROMPT.
- `[[ -o prompt_percent ]]` guard.
- Conditional wrap: only patch PROMPT when last in `precmd_functions`; otherwise self-reorder + `printf` mark1 once as fallback.
- `zle-line-init` fallback emits `133;P;k=i\a 133;B\a`.

**bash** (`etc/shell/ghostel.bash`)
- Inline marker in PS1 is now `133;P;k=i` (was 133;A). Readline redraws PS1 on every keystroke; embedding 133;A would fresh-line on every redraw.
- Newline injection: `133;P;k=s`. PS2 wrap: `133;P;k=s` + `133;B`.
- Emit `133;A;redraw=last;cl=line;aid=$BASHPID` once per cycle via `printf` at end of `__ghostel_wrapped_prompt_command` — fresh-line fires only at prompt-cycle start.

**fish** (`etc/shell/ghostel.fish`)
- Drop the `fish_prompt` wrap. Ghostty does not emit 133;B in fish (no "after fish_prompt" event); cells from 133;A→133;C are tagged `.prompt` in libghostty which is equivalent to `.input` for the link-skip logic.
- `__ghostel_mark_prompt_start` (on fish_prompt + fish_posterror), `__ghostel_mark_output_start` (on fish_preexec), `__ghostel_mark_output_end` (on fish_postexec).
- Detect fish 4.1+ for `click_events=1` parameter on 133;A.

**zig + elisp**
- `src/module.zig` OSC 133 dispatcher accepts `P` alongside A/B/C/D.
- `ghostel--osc133-marker` treats `P` the same as `A` for `ghostel--prompt-positions`.

OSC terminator switched to BEL (`\a`) everywhere — was ESC `\` on the zsh side. Both are valid; ghostty uses BEL consistently.

## Test plan
- [x] `make -j4 all test-evil` (220 elisp + 50 evil + zig + lint) — all green.
- [x] `make test-native` (117 native, including 2 new e2e zsh tests) — all green.
- [x] New `ghostel-test-zsh-line-init-fallback-no-fresh-line` spawns zsh, forces the fallback path, asserts cursor lands at col 8 (beside `final-> `) instead of col 0 of the next row. Verified to fail without the fix.
- [x] New `ghostel-test-zsh-prompt-cells-tagged-with-self-reorder-theme` spawns zsh, registers a self-reordering fake-theme that overrides PROMPT each cycle (mimicking `_p9k_precmd`), asserts `top-line` row carries `ghostel-prompt`.
- [ ] **Live verification on a real powerlevel10k session** — needs to be done by you; I don't have that setup locally. Should confirm cursor lands beside `❯ ` instead of below it.

### DIFF
--- a/etc/shell/ghostel.bash
+++ b/etc/shell/ghostel.bash
@@ -21,24 +21,41 @@ builtin command stty echo 2>/dev/null
 
 # Report working directory to the terminal via OSC 7
 __ghostel_osc7() {
-    printf '\e]7;file://%s%s\e\\' "$HOSTNAME" "$PWD"
+    printf '\e]7;file://%s%s\a' "$HOSTNAME" "$PWD"
 }
 
 # --- Semantic prompt markers (OSC 133) ---
+#
+# Marker layout mirrors ghostty's bash integration:
+#   - 133;A emitted via printf at end of PROMPT_COMMAND (once per
+#     cycle, with redraw=last;cl=line;aid=$BASHPID).
+#   - PS1 wrapped with 133;P;k=i (initial-prompt) at start and 133;B
+#     (input boundary) at end.
+#   - PS1 newlines (the bash `\n' escape) get 133;P;k=s injected.
+#   - PS2 wrapped with 133;P;k=s at start and 133;B at end.
+#
+# Why not embed 133;A inline in PS1?  133;A has fresh-line behavior
+# (CR+LF when cursor is not at column 0).  bash readline redraws PS1
+# on every keystroke that changes display state — embedding 133;A
+# would CR+LF on every redraw, eating the prompt.  133;P is the
+# side-effect-free equivalent.
+#
+# `\[ \]' mark the OSC sequence as zero-width for readline's line-wrap
+# math; `\a' is BEL — used everywhere here because `${var//pat/repl}'
+# eats backslashes in the replacement, which would break ST (`\e\\').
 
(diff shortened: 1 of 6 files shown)


dakra:
merged into main

Issue #246, opened 2026-05-09
Issues with evil single line movement/editing


noctuid:
Followup from https://github.com/dakra/ghostel/issues/218

Thanks a lot. It is working much better now. I still have some minor issues though:

- `word word word<esc>bbcw<esc>wdw` - first cw works, second deletes everything (I can *only* reproduce in zsh or pi for some reason, not bash)
- Commands like `evil-beginning-of-line` don't respect prompt boundaries (e.g. will enter the prompt text in bash or zsh, then `i` will move the point to after the prompt). I have also tried binding the keys I use to send home/end or c-a/c-e to the terminal, which works in insert state but not normal state when evil-ghostel-mode is enabled (have to bind it to enter insert and send)
- I am still seeing some very weird edge cases with operators *sometimes* affecting the wrong text, will update if I can give reliable reproduction steps

dakra:
You want to try #250 ?

For the key-binding problem in normal mode there is now `evil-ghostel-send-and-follow` that you should use to send keys when using evil in normal mode.

noctuid:
In zsh and pi `word word word<esc>bbcw` no longer works. It just enters insert and moves the point either to the beginning (zsh) or end (pi) of the prompt. This only happens the first time in pi (2nd time cw works) and not always in zsh. I'm also have some other strange issues where e.g. `w` no longer works and says it's at the end even though there is text after, and `dw` won't delete the last word either.

milanglacier:
I personally feel that using Evil mode motions within Emacs+terminal buffer is unnecessarily complex. 

Because a terminal buffer has a a lot of difference from a standard text buffer, applying vim-style text operations there can feel very brittle and hackisj.

Consequently, I prefer to rely on the shell's native Vim mode for text manipulation, as modern shells like Zsh and Nushell come fully equipped with this functionality.

noctuid:
> applying vim-style text operations there can feel very brittle and hackisj

Everything basic I use works fine in vterm, and even some very complex things like operators+avy work without issues. You can call it a hack, but it definitely doesn't *feel* like a hack.

> Consequently, I prefer to rely on the shell's native Vim mode for text manipulation, as modern shells like Zsh and Nushell come fully equipped with this functionality.

[...]

Issue #218, opened 2026-05-02
Evil multi-line issues in pi


noctuid:
Evil support does not appear to work with multi-line editing in pi (haven't tried other tui programs but could be the same).  For example, inserting a couple newlines, entering normal state and going up to the first line, then pressing dd will delete the last line not the current line.

dakra:
Can you try the PR #222 ?
I'm not an evil user so my testing was very limited.

noctuid:
Thank you, seems improved but still quite a lot of issues, e.g. `evil-open-below` and `evil-open-above` don't work correctly. Going up lines and then running `evil-append-line` puts the cursor at a later line. Same for `evil-insert-line`. `C` will delete the last line it seems as well.

dakra:
@Cianidos do you mind to have a look at this?

As I don't use evil it's always a bit of a pain for me to try to reproduce and fix those issues.

noctuid:
Multi-line editing is not a huge priority for me since I can compose longer prompts in a separate buffer and send them to ghostel. The single line editing also has some issues though, e.g. `evil-beginning-of-visual-line` and `evil-beginning-of-line` appear to not work in insert state (and the end equivalents).

dakra:
@noctuid ok. maybe you want to give it a shot for a PR? You can base it from my PR here if that helps (or not of course)

Cianidos:
I want to clarify the intended contract for `evil-ghostel` here.

@noctuid, do I understand your expectation correctly: normal-state editing commands such as `d`, `c`, `r`, `p`, etc. should work against the live terminal input by being translated into PTY key sequences?

@dakra, I noticed that the README currently documents this as supported behavior:

- delete/change/replace/paste are listed as normal-state operations handled by `evil-ghostel`
- insert/append sync the terminal cursor to point before entering insert state
- alt-screen programs are described as unaffected

Was that README section meant as the deliberate supported UX contract for `evil-ghostel`, or did it grow out of the implementation and maybe now needs to be reconsidered?

The reason I ask is that, in practice, this contract seems very hard to make robust for arbitrary terminal programs. Ghostel can move the terminal cursor only by sending key sequences to the PTY, but the running program owns the real input state. For shells/readline this may be manageable, but TUIs can have input fields, modal state, cursor constraints, wrapped prompts, padding, and application-specific key handling. In those cases, translating Evil buffer edits into PTY actions seems fragile and likely to have many corner cases.

[...]

PR #250, opened 2026-05-09
Fix evil motion crossing into stale whitespace cells (#246)

dakra:
## Summary

Fixes the operator misfires reported in #246 (and supersedes the diagnosis from #218):

- **#1 ('cw' then 'dw' over-deletes in zsh / pi)** — clamp evil's forward / end-of-line motions at the terminal cursor's buffer position so they can't walk into trailing whitespace cells left over from a non-CSI-K redraw.
- **#2a ('0' / '^' lands on the prompt without OSC 133)** — fall back to sending readline 'C-a' when the cursor's row carries no 'ghostel-prompt' property; '--sync-point-on-next-redraw' lands point at the new cursor.
- **#2b (custom normal-state passthrough binding doesn't move point)** — add public 'evil-ghostel-send-and-follow' as an opt-in wrapper for user bindings; doesn't change the redraw advice's default point preservation.
- **#3 (sporadic operator misfires)** — same root cause as #1; covered by the clamp.

## Mechanism

zsh's zle and prompt_toolkit redraw inline deletions by overwriting with spaces rather than emitting CSI K. The deleted columns end up as real space characters in the Emacs buffer past the live end of input. evil's 'evil-forward-word-begin' treats those cells as content and walks off the input row to the next buffer line; the subsequent 'dw' from the wrong row sends down-arrows through the PTY, which zsh interprets as history-next, scrubbing the input.

The clamp narrows the buffer's upper bound to the cursor's buffer position before the motion runs. When point is past the cursor (scrollback) the narrow is skipped so it can't cut point out of the visible region.

## Test plan

- [x] 'make -j4 all test-evil' — clean (lint, byte-compile, 85 evil-ghostel + 149 elisp + 331 native tests pass)
- [x] Live zsh repro (drives a real subprocess + libghostty pipeline): before — 'w' jumped to line 2, 'dw' triggered zsh history-jump; after — 'w' holds, 'dw' is a no-op
- [x] 7 new ERT tests covering the clamp, scrollback no-clamp, line-mode no-clamp, end-to-end cw->dw repro, 'C-a' fallback, scrollback fall-through, and 'send-and-follow'
- [ ] Final live verification in interactive emacsclient against noctuid's actual workflow

## Out of scope

- Reproducing #2b — not yet repro'd. The helper unblocks user-installed passthroughs without changing defaults; if their case isn't fully fixed by the helper we can iterate.
- Auto-sourcing the shell integration (the proper fix for #2a is OSC 133; the 'C-a' fallback is the safety net).

### DIFF
--- a/extensions/evil-ghostel/evil-ghostel.el
+++ b/extensions/evil-ghostel/evil-ghostel.el
@@ -145,6 +145,41 @@ helper is unsafe there — use `evil-ghostel--last-cursor-line' instead."
     (when cursor-line
       (= (- (line-number-at-pos (point) t) 1) cursor-line))))
 
+(defun evil-ghostel--cursor-buffer-position ()
+  "Return the buffer position at the terminal cursor's row+col, or nil.
+Translates `ghostel--cursor-pos' through the scrollback offset to a
+buffer line, then walks to the cursor's column.  Used to clamp
+forward motion at the end of input — anything past this position on
+the cursor's row is stale whitespace from a redraw that didn't emit
+`CSI K' (zsh's zle, prompt_toolkit at certain widths)."
+  (let ((line (evil-ghostel--cursor-buffer-line))
+        (cur ghostel--cursor-pos))
+    (when (and line cur)
+      (save-excursion
+        (goto-char (point-min))
+        (forward-line line)
+        (move-to-column (car cur))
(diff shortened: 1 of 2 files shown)


dakra:
This should be superseeded by #264

PR #253, opened 2026-05-10
Implement bold-color support (bright and fixed)

mgeisler:
Add support for [Ghostty's `bold-color` configuration](https://ghostty.org/docs/config/reference#bold-color), allowing bold text to be rendered using bright colors or a specific fixed color.

Summary of changes:
- src/terminal.zig: Add `BoldConfig` and `bold_config` field to `Terminal`.
- src/module.zig: Add `ghostel--set-bold-config` to update bold settings.
- src/render.zig: Implement color substitution logic in `readCellProps`. Bold text with palette colors 0-7 is mapped to 8-15. Bold text with default foreground uses `bold_config.fixed_color` if mode is `fixed`.
- lisp/ghostel.el: Add `ghostel-bold-color` defcustom (default `bright`) and logic to apply it to new and existing terminals.
- test/ghostel-test.el: Add unit tests for both 'bright and fixed-color bold modes.
- src/emacs.zig: Add `bright` symbol to cached symbols.

Matches Ghostty 1.2.0 behavior.

> Written by Gemini, tested by me.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -873,6 +873,7 @@ Used when `cursor-in-non-selected-windows' resolves to box.")
 (declare-function ghostel--mouse-event "ghostel-module")
 (declare-function ghostel--new "ghostel-module")
 (declare-function ghostel--redraw "ghostel-module" (term &optional full))
+(declare-function ghostel--set-bold-config "ghostel-module")
 (declare-function ghostel--set-default-colors "ghostel-module")
 (declare-function ghostel--set-palette "ghostel-module")
 (declare-function ghostel--set-size "ghostel-module" (term rows cols &optional cell-w cell-h))
@@ -1233,6 +1234,36 @@ Updated whenever the terminal is created or resized.")
 (defvar-local ghostel--term-cols nil
   "Column count of the native terminal.
 Updated whenever the terminal is created or resized.")
+
+(defcustom ghostel-bold-color nil
+  "Configure how bold text is colored.
+
+If nil (default), bold text uses the same color as normal text.
+
+If `bright', bold text uses the bright version of the current
+foreground color (ANSI colors 0-7 map to 8-15).
+
+If a string (hex color like \"#RRGGBB\"), bold text with the default
+foreground color uses this specific color.  Bold text with a palette
+color (0-7) will use the bright version (8-15).
+
+Matches Ghostty 1.2.0's `bold-color' configuration."
+  :type '(choice (const :tag "None" nil)
+                 (const :tag "Bright" bright)
+                 (string :tag "Fixed color (#RRGGBB)"
+                         :match (lambda (_widget val)
+                                  (and (stringp val)
+                                       (string-match-p
+                                        "\\`#[0-9A-Fa-f]\\{6\\}\\'" val)))))
+  :group 'ghostel
+  :set (lambda (sym val)
+         (set-default sym val)
+         (dolist (buf (buffer-list))
+           (with-current-buffer buf
+             (when (and (derived-mode-p 'ghostel-mode) ghostel--term)
+               (ghostel--apply-bold-config ghostel--term)
+               (let ((inhibit-read-only t))
+                 (ghostel--redraw ghostel--term t)))))))
 
 (defvar-local ghostel--cursor-pos nil
   "The position of the terminal cursor as (COL . ROW) in terminal screen coords.")
@@ -5023,6 +5054,15 @@ Falls back to \"#000000\" if the color cannot be resolved."
               "")))
         (ghostel--set-palette term colors)))))
 
+(defun ghostel--apply-bold-config (term)
+  "Apply `ghostel-bold-color' to terminal handle TERM."
+  (when (user-ptrp term)
+    (ghostel--load-module)
+    (ghostel--set-bold-config
+     term (if (eq ghostel-bold-color 'bright)
+              'bright
+            ghostel-bold-color))))
+
 
 ;;; Theme synchronization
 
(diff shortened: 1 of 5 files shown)


src/render.zig
```diff
@@ -75,7 +75,7 @@ fn formatColor(color: gt.ColorRgb, buf: *[7]u8) []const u8 {
 }
 
 /// Read the style for the current cell from the render state.
-fn readCellProps(cells: gt.RenderStateRowCells, key: CellPropKey) !?CellProps {
+fn readCellProps(term: *Terminal, cells: gt.RenderStateRowCells, key: CellPropKey) !?CellProps {
```
emil-e: There's a somewhat significant refactoring that's coming in the font glyph alignment PR that changes the structure a bit. In the new structure, the bold configuration should be a property of the `Renderer` rather than the `Terminal` since it's purely a rendering configuration.

lisp/ghostel.el
```diff
@@ -1109,6 +1112,32 @@ Updated whenever the terminal is created or resized.")
   "Column count of the native terminal.
 Updated whenever the terminal is created or resized.")
 
+(defcustom ghostel-bold-color 'bright
```
emil-e: I think it's good to achieve feature parity with Ghostty proper, but I'm thinking that maybe we should do this in a more Emacsy way using faces for customization instead.

@dakra What do you think?

dakra: If it's possible with defface, that would make more sense, but haven't thought about it too much.

But for sure personally I'm happy with just having a bold face and don't need the colors to be brighter,
so I think the default should be off (like it is in newer ghostty versions as well)

src/render.zig
mgeisler: Thanks, this has been refactored to use `Renderer`.

lisp/ghostel.el
mgeisler: I went with the simple solution for now with

```elisp
(defcustom ghostel-bold-color nil
  "Configure how bold text is colored.

If nil (default), bold text uses the same color as normal text.

If `bright', bold text uses the bright version of the current
foreground color (ANSI colors 0-7 map to 8-15).

If a string (hex color like \"#RRGGBB\"), bold text with the default
foreground color uses this specific color.  Bold text with a palette
color (0-7) will use the bright version (8-15).

Matches Ghostty 1.2.0's `bold-color' configuration."
```

> I'm thinking that maybe we should do this in a more Emacsy way using faces for customization instead.

I suppose one could accept a face instead and read its `:weight`? I'm not super sure how this should work, though since all (or nearly all?) other attributes of the face would be ignored.

I hope we can merge the simple approach first and then refactor later if there is a more elegant way.

emil-e:
I'll have another look tomorrow!

src/module.zig
```diff
@@ -1088,6 +1095,36 @@ fn fnSetDefaultColors(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value,
...
 
+/// (ghostel--set-bold-config TERM CONFIG)
+///
+/// CONFIG can be nil (none), 'bright, or a hex color string.
+fn fnSetBoldConfig(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value, _: ?*anyopaque) callconv(.c) c.emacs_value {
+    const env = emacs.Env.init(raw_env.?);
+    const term = env.getUserPtr(Terminal, args[0]) orelse return env.nil();
+    const val = args[1];
+
+    if (!env.isNotNil(val)) {
```
emil-e: There `isNil` for this

src/module.zig
```diff
@@ -1088,6 +1095,36 @@ fn fnSetDefaultColors(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value,
...
+        term.renderer.bold_config = .{ .mode = .none };
+    } else if (env.eq(val, emacs.sym.bright)) {
+        term.renderer.bold_config = .{ .mode = .bright };
+    } else {
+        const hex = env.extractStringAlloc(val, std.heap.c_allocator) orelse {
```
emil-e: Since the max size is know, you can use `extractString` instead.

src/Renderer.zig
```diff
@@ -38,6 +38,16 @@ row: RowContent = .{},
 
 font_info: ?FontInfo = null,
 
+/// Bold text coloring configuration.
+bold_config: BoldConfig = .{},
+
+pub const BoldConfig = struct {
```
emil-e: Since `fixed_color` is only valid if `mode` is `fixed`, use a tagged union (`union(enum)`) instead.

emil-e (changes requested):

src/Renderer.zig
mgeisler: Thanks, much better indeed.

emil-e (approved):

emil-e:
LGTM

@dakra Any thoughts? I personally don't quite understand the point but given Ghostty proper has this, I'm guessing it's a thing that people want? :)

dakra:
> LGTM

lgtm. Thanks, merging.

> @dakra Any thoughts? I personally don't quite understand the point but given Ghostty proper has this, I'm guessing it's a thing that people want? :)

Personally I'm also happy with just a bold face but I also don't have any aesthetic taste ;)
I asked the LLM why someone would want that option and the reasons make sense to me:

  - **Legacy/theme convention**. Old terminals couldn't render bold glyphs, so "bold" historically meant "switch to
  the bright palette slot." Most color schemes (Solarized, Gruvbox, Nord, etc.) were tuned assuming bold red =
  bright red — without the brightening, output from ls, grep, prompts, etc. looks duller than the theme author
  intended.
  - **Visual punch**. Bold-weight alone can be subtle at small sizes or with thin fonts; the color shift makes
  emphasized tokens actually pop.
  - **Faux-bold avoidance**. Some users disable bold-weight (eyestrain, ligature issues, fonts without a real bold
  face) but still want emphasis — bold-color = "#hex" gives them a single accent color instead of synthetic stroke
   widening.

PR #264, opened 2026-05-11
Rewrite evil-ghostel from advice to command-remap architecture

dakra:
## Summary

Rewrites evil-ghostel from an advice-based design to proper evil command definitions, and makes input-boundary detection autosuggestion-aware.

- Replaces ~13 `advice-add` hooks on `evil-*` commands with `evil-define-operator` / `evil-define-motion` definitions bound in `evil-ghostel-mode-map` for normal and visual states (vterm-collection style). Drops the shadow-cursor model — operators read `ghostel--cursor-pos` directly each call.
- **Operator clamp** trims an operator's range to the live input region, so `dw` on the last input word no longer over-deletes into the blank renderer rows below the prompt. Forward-word motions stay vanilla in operator-pending state so `dw` / `cw` can overshoot and get trimmed by the clamp.
- **Autosuggestion-aware boundaries.** Shell suggestions (zsh-autosuggestions, fish) are tagged `ghostel-input` like typed text and differ only by a dimmer `:foreground`. `evil-ghostel--suggestion-p` / `--input-end` keep `bdw` / `bcw` / `dw` / `^` / `$` from running into the grey suggestion, and never return a position left of the cursor (fixes `A` / `$` / `d$` driving a left-arrow into the prompt on an empty prompt).
- **Visual selection works.** `evil-ghostel-mode` opts the buffer out of ghostel's keyboard-mark→copy-mode switch (buffer-local `ghostel-mark-activation-input-mode` nil), so `v` and the visual operators no longer flip the buffer to read-only copy mode and hit "Buffer is read-only". Mouse selection stays governed by `ghostel-mouse-drag-input-mode`.
- Keeps the essential plumbing advice (`ghostel--redraw` / `ghostel--set-cursor-style`) and the insert-state-entry hook, installed on first mode-enable and removed only when the **last** `evil-ghostel-mode` buffer disables.
- Three-way insert-state ESC routing (`evil-ghostel-escape`: `auto` / `terminal` / `evil`), toggleable per buffer.
- `evil-ghostel-goto-input-position` drives the terminal cursor with arrow keys, including vterm-style recovery for literal `^[[C` echo and bash-autosuggest accept-on-right-arrow.

## Fixes (#264 reports)

- fish's right-aligned prompt cells — which libghostty's per-cell heuristic tags `SEMANTIC_INPUT` despite emitting no OSC 133;B — are no longer captured by `i` / `a` / `I` / `A` / `$`.
- `r` reads its replacement via `evil-read-key` (the prior `<c>` interactive code captured COUNT, not the char, so `r` was a no-op).
- `cw` → `ce` via `evil-change-commands`; delete count excludes soft-wrap newlines.

## Test plan

- [x] `make -j8 all test-evil` clean; **107** evil-ghostel tests, 0 unexpected.
- [x] New regression coverage for suggestion-aware boundaries, empty-prompt left-arrow, the advice lifecycle, and the `--sync-render` forced-redraw path.
- [x] Live-verified in a sandboxed live Emacs against `zsh`: `v` stays in semi-char and editable (no read-only freeze), `^ v l l d` deletes input over the PTY, and suggestion-aware boundaries hold.

## Notes for reviewers

- The `evil-ghostel.el` size delta vs evil-collection-vterm is dominated by `evil-ghostel--around-redraw` (ghostel's renderer wipes the viewport region, vterm doesn't), the cursor↔point viewport-row sync helpers, the insert-state-entry hook (vterm's cursor *is* the buffer cursor), and the three-way ESC routing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,7 @@ jobs:
           emacs --batch -Q -L /tmp/compat -L /tmp/evil -L lisp -L extensions/evil-ghostel \
             -l ert \
             -l test/evil-ghostel-test.el \
-            -f evil-ghostel-test-run-elisp
+            -f evil-ghostel-test-run
 
   build-native:
     strategy:
--- a/extensions/evil-ghostel/evil-ghostel.el
+++ b/extensions/evil-ghostel/evil-ghostel.el
@@ -12,9 +12,20 @@
 
 ;;; Commentary:
 
-;; Provides evil-mode compatibility for the ghostel terminal emulator.
-;; Synchronizes the terminal cursor with Emacs point during evil state
-;; transitions so that normal-mode navigation works correctly.
+;; Evil-mode integration for the ghostel terminal emulator, modeled on
+;; `evil-collection-vterm'.
+;;
+;; Like vterm's, the surface area is tiny.  Basic motions (h l w b e $ 0)
+;; stay vanilla Evil, so point moves freely over the rendered row.  Only
+;; the commands that drive the shell's line editor are customized:
+;; operators (d c x s r and line variants) clamp their range to
+;; [input-start, input-end] and apply it over the PTY (arrows, backspaces,
+;; bracketed paste); i a I A drive the shell cursor to point; ^ jumps past
+;; the prompt; j / G stay on the live prompt; p / P paste; u sends
+;; readline undo.
+;;
+;; Outside semi-char input mode (`line' / `copy' / `emacs' / `char' modes,
+;; or an alt-screen TUI) every command falls through to `evil-*'.
 ;;
 ;; Enable by adding to your init:
 ;;
@@ -41,11 +52,8 @@
 
 (defcustom evil-ghostel-initial-state 'insert
   "Initial evil state for new `ghostel-mode' buffers.
-Setting this option via `customize-set-variable', `setopt', or the
-Customize UI calls `evil-set-initial-state' so the change takes effect
-immediately.  Users who prefer the raw API can call
-`evil-set-initial-state' directly from their config — the registry is
-last-writer-wins."
+Setting via Customize, `setopt', or `customize-set-variable' applies the
+change immediately through `evil-set-initial-state'."
   :type '(choice (const :tag "Emacs" emacs)
                  (const :tag "Insert" insert)
                  (const :tag "Normal" normal)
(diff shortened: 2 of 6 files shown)


noctuid:
@dakra Thanks a lot. I didn't have time to test thoroughly. I'll test tomorrow but couldn't find any major problems with basic use. Here are two things:

- it should use remaps only, not bind any specific keys
- minor: `word<esc>a` (`evil-ghostel-append`) will add a visual space that isn't there (point will be after a space but backspace in insert would delete the "d"); test with `evil-move-cursor-back nil` and zsh (not sure if this is specific to my config, couldn't reproduce in bash)

dakra:
@noctuid Thanks. I fixed the 2 bugs you mentioned.
I'll still have to have a closer look at it and test it more. but if you want to give it another test ride that's obviously helpful as well.

noctuid:
I still haven't seen any more issues with editing/operators, though I'm seeing similar issues to the appending one. Sometimes the cursor will jump to the right side of the window when appending.

`evil-forward-char` can sometimes go many characters beyond the end of the text in the prompt line in normal state (e.g. in pi always and in zsh I can go right as far as I had previously typed text even if it was deleted). Same applies to e.g. 0/$.

I saw weird point warping a couple of times but can't reliably reproduce.

AndiLavera:
Tested on aarch64-darwin, emacs 30, doom/evil latest with fish. Basic nav seems to be working. 

I have the git branch justified right like:
```fish
~/.c/doom ❯❯❯ nslookup -type=txt on.quad9.net 2620:fe::9                                      main ✱
```

- `I` will place the insert cursor perfectly before the `nslookup`. `A` will place the insert cursor after the `main ✱` until first keystroke.
- `$` places my cursor on the `✱` in `main ✱`
- `y$` will copy the `main ✱`

When I move out of a git repo, the `main *` in my terminal is gone:
```fish
~ ❯❯❯ nslookup -type=txt on.quad9.net 2620:fe::9
```

These bugs do not persist. `A` snaps directly after the `9`, `$` places the cursor on the `9`, `y$` copies only the written text and does not contain whitespace characters after the `9`

iftheshoefritz:
Is this intended to fix insert at cursor problems as well? Currently:

`A 1234 ESC h h i 1` inserts the `1` at the end of the line after `4` instead of just after the first `1`.

Also given `☁  ~  12341`, `^` takes me to the cloud in my prompt, not to the first character I have control over (`1`).

dakra:
I experimented a bit with different versions today.
It's very fragile, I feel like every time I fixed one bug, another edge-case triggered and didn't work :(
Especially fish with a right_prompt was tricky.

But in my short testing it seems to work now.
Can you test again @AndiLavera @iftheshoefritz @noctuid ?

noctuid:
@dakra Is this PR being canceled or replaced?

dakra:
@noctuid oh. I was doing some housekeeping in magit of git-refs and deleted a bunch of old branches.
This one was not supposed to be in there, sorry 🙈

iftheshoefritz:
None of these are fixed:

```
A 1234 ESC h h i 1 inserts the 1 at the end of the line after 4 instead of just after the first 1.

Also given ☁  ~  12341, ^ takes me to the cloud in my prompt, not to the first character I have control over (1).
```

iftheshoefritz:
Also a weird one: typing `f` flickers the cursor briefly to the position just before as the letter appears. This doesn't happen for any other character in the (English) alphabet.

dakra:
> A 1234 ESC h h i 1 inserts the 1 at the end of the line after 4 instead of just after the first 1.

that works for me.
Do you have the latest version? I rebased, so you have to do a git hard reset.

> Also given ☁  ~  12341, ^ takes me to the cloud in my prompt, not to the first character I have control over (1).

Do you have shell integration on?
What shell are you using?

dakra:
@iftheshoefritz did you see my last comment?

Wondering if I should merge this? This non-advice version looks like it's improved but I don't want to break anything.

dakra:
Also tagging @AndiLavera @noctuid @Cianidos to maybe give feedback.

Sorry if it's annoying, I just want to make sure I don't merge a version that's actually worse than what we currently have :D

noctuid:
I'm still have strange issues, e.g. `word word<esc>bdw` or `word word<esc>bC` deletes both words

noctuid:
Above issue is in zsh. Some other things: 

- in bash, `word word<esc>bcw` (or dw)  also deletes the space before the word
- `^` does visually put the cursor before the prompt, cursor location is corrected after `i`
- `$` - would expect to put one char to the right of where it actually puts the point; or sometimes it puts the point way over to the right visually past where it really is
- `evil-ghostel-backward-char` is getting stuck at the end of lines for me in zsh, always moves to the end of the input (works in bash)

dakra:
@iftheshoefritz @AndiLavera @noctuid @Cianidos sorry to bother everyone again but I did a lot of testing.
I even wrote my own tool to help me automate the evil tests for different scenarios in bash, fish, zsh and a python repl.

Would you mind giving this PR another try and give feedback?
It's also rebased on the latest main branch so you're not missing out on the new features :)

dakra:
Merged this now into main (the PR is already over  a month old).
Hope it's working good for everyone.
Please open a issue if you find any problems or have new feature suggestions.

Issue #47, opened 2026-04-06
Make buffer creation behave like other packages?


affe-mit-waffel:
Hi there,

while working on [this pull requests](https://github.com/dakra/ghostel/pull/45) I was thinking about how `ghostel` creates new terminal buffers. Everytime the user executes `M-x ghostel` a new unique buffer is created (and the count of `ghostel--buffer-counter` is increased). 

This is quite different from  eshell, term, vterm and eat: They switch to the existing buffer when you run the command for a second time. Eshell for example accepts an universal argument to force the creation of a second unique buffer.
Because if my muscle memory I was ending up with a high count of ghostel buffers.

Is this something you would consider changing making it more similar to existing packages? Or is this a very deliberate design choice. That would be fine too!

Anyways: Thanks for your hard work! 😊

affe-mit-waffel:
Nice, I will prepare a merge request

PR #266, opened 2026-05-11
Bind `Shift-PageUp`/`Down` to `scroll-down`/`up-command`

mgeisler:
This change allows users to use `Shift-PageUp` and `Shift-PageDown` to scroll through terminal scrollback in `semi-char` mode, matching the behavior of traditional terminal emulators.

- Add `S-<prior>` and `S-<next>` to `ghostel-keymap-exceptions`.
- Bind `S-<prior>` and `S-<next>` in `ghostel-mode-map`.
- Add regression tests for `semi-char` and `char` modes.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -634,7 +634,7 @@ When nil, falls back to `tramp-default-method'."
                  string))
 
 (defcustom ghostel-keymap-exceptions
-  '("C-c" "C-x" "C-u" "C-h" "M-x" "M-o" "M-:" "C-\\")
+  '("C-c" "C-x" "C-u" "C-h" "M-x" "M-o" "M-:" "C-\\" "S-<prior>" "S-<next>")
   "Key sequences that should not be sent to the terminal.
 These keys pass through to Emacs instead."
   :type '(repeat string))
@@ -1560,6 +1560,9 @@ Input modes (`ghostel-semi-char-mode-map', `ghostel-char-mode-map',
   "C-c C-y"          #'ghostel-paste
   "C-c M-l"          #'ghostel-clear-scrollback
   "C-c C-q"          #'ghostel-send-next-key
+  ;; Scrolling
+  "S-<prior>"        #'scroll-down-command
+  "S-<next>"         #'scroll-up-command
   ;; Hyperlink navigation (OSC 8, auto-detected URLs, file:line refs)
   "C-c C-n"          #'ghostel-next-hyperlink
   "C-c C-p"          #'ghostel-previous-hyperlink
--- a/test/ghostel-test.el
+++ b/test/ghostel-test.el
@@ -10173,6 +10173,22 @@ instead of running Emacs commands like `forward-sexp'."
           (should-not (eq binding #'ghostel--send-event))
         (should (eq binding #'ghostel--send-event))))))
 
+(ert-deftest ghostel-test-shift-page-up-down-bindings ()
+  "S-<prior> and S-<next> should be bound to scrolling commands.
+They are excluded from terminal-key definitions by `ghostel-keymap-exceptions'
+and inherited from `ghostel-mode-map'."
+  (should (eq (lookup-key ghostel-mode-map (kbd "S-<prior>")) #'scroll-down-command))
+  (should (eq (lookup-key ghostel-mode-map (kbd "S-<next>")) #'scroll-up-command))
+  ;; Inherited in semi-char mode
+  (should (eq (lookup-key ghostel-semi-char-mode-map (kbd "S-<prior>")) #'scroll-down-command))
+  (should (eq (lookup-key ghostel-semi-char-mode-map (kbd "S-<next>")) #'scroll-up-command)))
+
+(ert-deftest ghostel-test-char-mode-shift-page-up-down-bindings ()
+  "In char-mode, S-<prior> and S-<next> SHOULD be sent to the terminal.
+Char mode ignores exceptions and binds everything to `ghostel--send-event'."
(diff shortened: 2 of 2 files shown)


dakra:
I think personally I'm against merging this.
My reasons would be:
- we should lean to how Emacs behaves normally and not an external terminal
- S-PageUp/Down are real CSI sequences and vim, tmux etc read them
- it's basically 2 lines for a user who wants it personally:

```
(use-package ghostel
  :bind (:map ghostel-semi-char-mode-map
         ("S-<prior>" . scroll-down-command)
         ("S-<next>"  . scroll-up-command)
[...]
```

but that is only my opinion of course, you can make the same argument the other way around that it's only 2 lines of custom config if you do NOT want the binding ;)

@emil-e you're the tiebreaker ;)
(or anyone else who is reading this can chime in of course)

emil-e:
@dakra I think I agree with your reasoning. I could be convinced otherwise, but I don't think it's obvious that this should be a default.

mgeisler:
> we should lean to how Emacs behaves normally and not an external terminal

I should expand on this, since I think I might be using and thinking of Ghostel differently than you. I say this because of how `M-x ghostel` returns to the same terminal buffer from anywhere.

I've been a long time Linux user and have recently started using a MacBook at work, against my better judgement :smile: So to bring some regularity back to my life, I started looking into using Emacs terminals to have a consistent workflow.

I spawn terminals often with a keyboard shortcut, so I have written a little helper for this:

```elisp
(defun mg/ghostel-new-frame ()
  "Create a new, dedicated frame for a Ghostel session.

This function creates a new frame without a menu bar, starts a
new, unique Ghostel session in the user's home directory, and
ensures the frame closes automatically when the shell exits, and
vice-versa: the buffer is killed if the frame is closed manually.

Run this with

  emacsclient --no-wait --eval \"(mg/ghostel-new-frame)\"

in your window manager."
  (interactive)
  (let ((ghostel-frame (make-frame '((name . "Ghostel")
                                   (menu-bar-lines . 0)))))
    (with-selected-frame ghostel-frame
      (let ((default-directory (expand-file-name "~")))
        (with-current-buffer (ghostel t)
          (let ((buffer (current-buffer)))
            ;; Direction 1: Kill frame when buffer is killed (shell exit)
            (add-hook 'kill-buffer-hook
                      `(lambda ()
                         (when (frame-live-p ,ghostel-frame)
                           (delete-frame ,ghostel-frame)))
                      nil 'local)
            ;; Direction 2: Kill buffer when frame is deleted
            (add-hook 'delete-frame-functions
                      (lambda (frame)
                        (when (eq frame ghostel-frame)
                          (kill-buffer buffer)))
                      nil 'local)))))))
```

Basically, I wish to get a new Emacs frame dedicated to the new terminal session. When it goes away, the frame goes away, and vice versa.

This is obviously very different from how `M-x ghostel` invites you to use it.

For me, this works almost completely, except for the 2 lines of extra space at the bottom (mode line and echo area). A bit of lost vertical space in those frames, but I can live with that.

Because I treat these windows as 95% a terminal, and 5% an Emacs frame, I instinctively reach for terminal shortcuts like Shift-PageUp etc. I also double-click text there, to quickly mark and copy it, just like in a terminal.

I somehow don't do this in a "regular" Emacs buffer, which I admit is a little weird :slightly_smiling_face: 

> * it's basically 2 lines for a user who wants it personally:
> 
> ```
> (use-package ghostel
>   :bind (:map ghostel-semi-char-mode-map
>          ("S-<prior>" . scroll-down-command)
>          ("S-<next>"  . scroll-up-command)
> [...]
> ```

Ah, well that is easy :slightly_smiling_face: In that case, the PR is of course not strictly needed. It then becomes more a question of what you are emulating, as you also note.

dakra:
> I should expand on this, since I think I might be using and thinking of Ghostel differently than you. I say this because of how M-x ghostel returns to the same terminal buffer from anywhere.

Same. I almost always call ghostel with a prefix arg.
I had it "my way" in the very beginning but it was "fixed" to be more in alignement with the other terminal packages in the very beginning. See https://github.com/dakra/ghostel/issues/47 

> I've been a long time Linux user and have recently started using a MacBook at work, against my better judgement 😄 

Same. Miss my Linux Thinkpad with i3.

> Ah, well that is easy 🙂 In that case, the PR is of course not strictly needed. It then becomes more a question of what you are emulating, as you also note.

jep. And for now I would say we leave it to be more close to what Emacs terminals do.
So far we mainly made special cases for prefix maps like C-c, M-x etc and paste commands because they would
be broken otherwise.

I'll close this then. If more people will vote for this we can change the default, I don't have a strong opinion.

mgeisler:
> Same. I almost always call ghostel with a prefix arg.
> I had it "my way" in the very beginning but it was "fixed" to be more in alignement with the other terminal packages in the very beginning. See #47

Ah, funny! Thanks for explaining the background.


> I'll close this then. If more people will vote for this we can change the default, I don't have a strong opinion.

Now that you documented the very simple solution (which I should have thought of myself), then I think it's all good.

If more similar cases pop up, then perhaps one could make some customize variables for toggles convenience.

For me, discovery would be helped with that: I as a user, don't really care about the terminal escapes sent when I press so and so — I care about `enable-traditonal-scrollback`. So having a Boolean I can toggle would let me both discover the concept and "export" a few lines of complexity.

dakra:
> So having a Boolean I can toggle would let me both discover the concept and "export" a few lines of complexity.

A boolean only for this obviously wouldn't make too much sense as it's 1 vs 2 line. but I agree with you that
maybe a mode that sits between semi-char-mode and line-mode.
Mostly semi-char but map a few more things to Emacs keybindings, like `M-v` could also scroll and `C-p` could move up tine the scrollback or something like that.

PR #279, opened 2026-05-16
Bind C-Backspace to send C-w in semi-char mode

lukejewers:
## Summary

Adds a `C-<backspace>` binding in semi-char mode that sends `C-w` to the terminal.

This gives shell prompts/readline-style editors the common 'kill previous word' behaviour when pressing `C-Backspace` in ghostel. Vterm, for example, has this behaviour by default.

No worries if this is considered too opinionated; I can keep it in my local config.

Ps. Thanks for making ghostel. It works great and has replaced my vterm/shell usage.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1815,7 +1815,8 @@ Most keys are sent to the terminal.  Keys in
   "C-y"            #'ghostel-yank
   "S-<insert>"     #'ghostel-yank
   "<remap> <yank>" #'ghostel-yank
-  "M-y"            #'ghostel-yank-pop)
+  "M-y"            #'ghostel-yank-pop
+  "C-<backspace>"  #'ghostel-backward-kill-word)
 
 ;; No parent — char mode captures everything, including C-c.
 (defvar-keymap ghostel-char-mode-map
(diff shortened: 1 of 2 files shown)


dakra:
Honestly, as long as not multiple users request it, I would prefer to keep ghostel clean from any opinionated settings.

E.g. this is currently my ghostel config: https://github.com/dakra/dmacs/blob/8b01aae2279501cf79a990e2348d628e13a7e1f9/init.el#L748-L792

There you see

```
         :map ghostel-semi-char-mode-map
         ("C-s"  . consult-line)
         ("C-k"  . ghostel-send-C-k-and-kill)
         ("M-d"  . ghostel-send-M-d-and-kill)
         ("M-<backspace>" . ghostel-backward-kill-word)

[...]

  (defun ghostel-backward-kill-word ()
    "Send `M-backspace' to ghostel.
Like normal Emacs `M-d'.  Kill a word backwards and put content in kill-ring."
    (interactive)
    (kill-ring-save (save-excursion (backward-word) (point)) (point))
    (ghostel-send-key "backspace" "alt"))
```

where I like for `C-k` and `M-<backspace>` to also kill whatever is deleted.

But someone else might have it to a completely different command.
So by default, I think pressing `C-<backspace>` should send exactly that to
the terminal and the running app then can decide how to handle it.

It might me a readline .inputrc config or some tui app config, but
I would think it's confusion if e.g. you configure readline to do X on
`C-<backspace>` and then not have it working until you figure out that actually
`C-w` is send to the terminal instead.

I'm open for discussion and to be convinced otherwise but as of right now
I lean towards declining this PR.
(Thanks of course anyway, even if I don't merge ❤️ )

lukejewers:
That’s fair enough. My thinking was that `C-<backspace>` conventionally means `backward-kill-word` in Emacs, and there is precedent for this behaviour in vterm. (Incidentally, I like your kill-ring aware version).

However, I see your point that translating `C-<backspace>` to `C-w` by default could be confusing, especially for users who expect the terminal application to receive the actual key sequence.

I’ll close the PR and keep this in my local config. Thanks for taking a look!

PR #284, opened 2026-05-17
Remove fnDebugState, fnDebugFeed and dead render state fields

emil-e:
Removes the two test-only debug functions that clobber render state, and the dead render_state/row_iterator/row_cells fields they were the sole consumers of.

## Zig changes
- Deleted `fnDebugState` (`ghostel--debug-state`) and `fnDebugFeed` (`ghostel--debug-feed`) from module.zig
- Removed `render_state`, `row_iterator`, `row_cells` fields from the Terminal struct (terminal.zig)
- Removed their init/error-path-cleanup/deinit code

## Test migration
- `ghostel-test--cursor`: now calls `ghostel--redraw` and reads the existing buffer-local `ghostel--cursor-pos` instead of parsing `ghostel--debug-feed` output
- `ghostel-test--row0`: now extracts the first line from `ghostel--copy-all-text` (formatter API, side-effect-free) instead of regex-parsing `ghostel--debug-state` output
- All 16 direct `ghostel--debug-state` call sites replaced with `ghostel--copy-all-text`

## Result
No more render-state-clobbering debug functions. All 180 native module tests pass.


emil-e:
@claude Review this

dakra:
@emil-e lgtm. Merge whenever you want 🚀

PR #286, opened 2026-05-18
Promote characters to wide if they have space available after

emil-e:
This PR improves glyph rendering by allowing characters to dynamically expand into adjacent empty space, rather than always being squashed to fit their allocated cell width.

## What changed

Previously, a glyph was scaled to fit either 1 or 2 cells depending on whether it was marked as wide. Now, if a glyph is wider (in aspect ratio) than its allocated slot, the renderer checks if the following characters are spaces. If so, it shrinks those spaces to zero-width via Emacs display properties, effectively claiming their cell width for the glyph while maintaining monospace alignment.

## Key details

- Lines are right-trimmed of trailing spaces, so positions at and past the newline represent empty space that can be freely claimed.
- On a genuinely full line, the column check prevents expansion before reaching the row boundary.
- Uses Emacs display properties (specifically `(space :width 0)`) to reclaim space without mutating buffer text.
- The aspect-ratio comparison ensures we only expand when the glyph is proportionally wider than the available slot.

## Files changed

- src/Renderer.zig — added dynamic width expansion logic and tracking of column position in CellInfo
- src/emacs.zig — added `char-after`, `space`, and `:width` to interned symbols


michaelneuper:
When trying to build I get some compiler errors: https://pastebin.com/fp0VtvDJ

dakra:
@michaelneuper you have to use zig 15.2.

as I see you use homebrew:
```
brew uninstall zig
brew install zig@0.15
```

emil-e:
@claude Review this

PR #292, opened 2026-05-19
Cache font metrics during redraw to reduce Unicode TUI stalls

### LINKED ISSUE #291: Unicode-heavy TUI redraws can freeze Emacs due to repeated font metric queries and GC churn
# Ghostel: Unicode TUI redraw freezes Emacs due to font metric hot path and GC churn

## Summary

Running a Unicode-heavy TUI such as `btop` inside Ghostel can make Emacs freeze
or sit at 100% CPU. The freeze is reproducible when `btop` updates quickly
(`100ms` update interval was enough on the affected setup). Profiling shows the
hot path is Ghostel redraw, specifically native renderer callbacks into Emacs
font APIs for glyph adjustment, amplified by frequent automatic GC.

## Environment

- OS: Linux
- Emacs: 30.2, graphical frame
- Config: Doom Emacs
- Ghostel: 0.27.0, straight checkout at `~/.emacs.d/.local/straight/repos/ghostel`
- Font: Iosevka, reported frame font `-UKWN-Iosevka-light-normal-normal-*-32-*-*-*-d-0-iso10646-1`
- Example terminal size during repro: approximately `238` columns, `29-50` rows
- TUI: `btop`, update interval `100ms`
- GCMH active: `gcmh-low-cons-threshold = 800000`, `gcmh-high-cons-threshold = 67108864`

## Reproduction

1. Start Emacs with Ghostel available.
2. Open a Ghostel buffer.
3. Run `btop` inside Ghostel.
4. Set `btop` update interval to `100ms`.
5. Let the TUI redraw continuously.
6. Observe Emacs responsiveness and CPU usage.

## Expected Behavior

Ghostel should remain interactive while a TUI redraws at a moderate refresh rate.
CPU usage may rise, but Emacs should not freeze or spend most time in GC/font
metric callbacks.

## Actual Behavior

Emacs becomes sluggish or freezes, with the Emacs process consuming high CPU.
Interactive commands and redisplay can be delayed significantly.

## Profiled Hot Path

CPU profiler sample during the freeze:

```text
Automatic GC                                                72%
server-process-filter                                       27%
  server--process-filter-all-pending
    server--process-filter-1
      server-execute-continuation
        accept-process-output
          timer-event-handler
            ghostel--delayed-redraw
              ghostel--redraw
                my/ghostel-preserve-evil-point-a
                  native module renderer
                    query-font                              26%
                    font-get-glyphs
                    font-at
```

A counter sample over a 2s window, before mitigation, showed a single Ghostel
redraw causing approximately:

- `2564` calls to `font-at`
- `2564` calls to `query-font`
- `2564` calls to `font-get-glyphs`
- `16` automatic GCs taking `1.83s` total during the sample

This makes the renderer/font path and GC dominate wall time.

## Suspected Cause

In `src/Renderer.zig`, Ghostel marks cells for glyph adjustment when a grapheme
cluster is present or when the codepoint is greater than or equal to a default
font coverage threshold:

```zig
if (self.graphemes.items.len > 1 or self.graphemes.items[0] >= adjustment_threshold) {
    try self.adjust_cells.append(...);
}
```

For every adjusted cell, `adjustGlyph` calls back into Emacs:

```zig
const font = env.f("font-at", .{ start_val, window });
const font_info = env.f("query-font", .{font});
const glyphs = env.f("font-get-glyphs", .{ font, start_val, end_val });
```

This is expensive for Unicode-heavy terminal frames. `btop` uses many box,
block, and braille graph characters, so each frame can enqueue thousands of
adjustment cells. On the affected system, the default Iosevka font already
reports coverage for representative btop glyph ranges:

- Box drawing `U+2500`: covered
- Block element `U+2588`: covered
- Braille `U+2800..U+28FF`: covered

So many fallback/glyph-adjustment checks appear to be conservative false
positives rather than actual font fallback cases.

## Workaround Tested in User Config

A Doom config workaround made the repro responsive immediately:

- Cache `query-font` results dynamically for the duration of one
  `ghostel--redraw` call.
- Temporarily raise `gc-cons-threshold` during Ghostel redraw only.

After this, the same `btop` session remained responsive even at a `100ms`
update interval.

## Proposed Fix Direction

Potential upstream fixes:

- Add per-redraw font metric caching in Ghostel so repeated `query-font` calls
  for the same font are not repeated thousands of times per frame.
- Cache glyph adjustment decisions/metrics in the native renderer, keyed by
  font and character/grapheme where safe.
- Refine the adjustment predicate so default-font-covered Unicode characters
  do not go through fallback adjustment unnecessarily.
- Optionally add a user-facing switch to disable or reduce expensive glyph
  adjustment for users who prefer terminal redraw performance over perfect
  fallback glyph fitting.

## Follow-Up Finding

An attempted native fast path that skipped glyph adjustment for single-width
characters rendered with the default font made btop fast but reintroduced visual
artifacts: shifted borders and glitchy TUI lines. That suggests Ghostel's
current glyph adjustment is still needed even when Emacs chooses the default
font for Unicode box/block/braille characters.

The safe patch direction is therefore narrower:

- Keep the existing glyph adjustment behavior.
- Cache `query-font` results for the duration of one redraw.
- Raise GC thresholds dynamically only around native redraw.

A fresh Emacs session with that narrower patch loaded showed no visible btop
artifacts, while a 2s sample dropped `query-font` calls to single digits and
kept GC time low.

## Notes

The issue is not caused by Evil point preservation advice in the local config.
During btop's alt-screen mode, that advice falls through to the original
`ghostel--redraw`; the profiler showed the cost inside the native renderer's
font metric calls and automatic GC.

Cianidos:
## Summary

This reduces redraw stalls for Unicode-heavy TUIs such as `btop` by avoiding repeated `query-font` calls for the same font during a single Ghostel redraw.

Fixes dakra/ghostel#291

## Changes

- Add `ghostel--query-font-cached`, dynamically backed by a redraw-local hash table.
- Bind the cache around the native renderer call in `ghostel--delayed-redraw`.
- Have the native renderer call `ghostel--query-font-cached` instead of `query-font`.
- Dynamically raise `gc-cons-threshold` and `gc-cons-percentage` only while rendering a frame.
- Add a focused test for the `query-font` cache.

## Why

Profiling a `btop` session showed redraw spending significant time in repeated Emacs font metric queries and automatic GC. A single redraw could call `query-font` thousands of times for repeated fonts.

This patch keeps existing glyph adjustment behavior intact. It intentionally does not skip `font-at` or `font-get-glyphs`, because an attempted default-font fast path caused visible TUI border artifacts.

## Validation

- `zig fmt src/Renderer.zig src/emacs.zig`
- `make .build/tests/elisp-ghostel-glyph-kitty-test.ok`
- `make .build/tests/native-ghostel-glyph-kitty-test.ok`

Manual validation:

- `btop` at `100ms` update interval remains responsive.
- No shifted border or visual artifacts observed after removing the unsafe glyph-adjustment skip.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1494,6 +1494,9 @@ Matches Ghostty 1.2.0's `bold-color' configuration."
 (defvar-local ghostel--rendered-font nil
   "The font last used for rendering. Internally used by native code.")
 
+(defvar ghostel--query-font-cache nil
+  "Dynamically bound cache for `query-font' during one native redraw.")
+
 (defvar-local ghostel--input-mode 'semi-char
   "Current input mode.
 One of `semi-char', `char', `copy', `emacs', or `line'.  See
@@ -6157,6 +6160,14 @@ frame after idle to improve interactive responsiveness."
                             #'ghostel--delayed-redraw
                             (current-buffer))))))
 
+(defun ghostel--query-font-cached (font)
+  "Return `query-font' metrics for FONT, caching during native redraw.
+When `ghostel--query-font-cache' is nil, call `query-font' directly."
+  (if ghostel--query-font-cache
+      (or (gethash font ghostel--query-font-cache)
+          (puthash font (query-font font) ghostel--query-font-cache))
+    (query-font font)))
+
 (defconst ghostel--line-context-lines 3
   "Number of lines (including the target) used as a disambiguation key.
 `ghostel--line-key' captures this many consecutive lines starting at a
@@ -6482,10 +6493,14 @@ new output arrives."
                        (ghostel--line-mode-snapshot)))
                  (inhibit-read-only t)
                  (inhibit-redisplay t)
-                 (inhibit-modification-hooks t))
+                 (inhibit-modification-hooks t)
+                 ;; Raise GC threshold to defer GC during redraw.
+                 (gc-cons-threshold (min most-positive-fixnum
+                                         (* gc-cons-threshold 3))))
             (when render-win
-              (with-selected-window render-win
-                (ghostel--redraw ghostel--term ghostel-full-redraw)))
+              (let ((ghostel--query-font-cache (make-hash-table :test 'eq)))
+                (with-selected-window render-win
+                  (ghostel--redraw ghostel--term ghostel-full-redraw))))
             (let ((line-restored
                    (and line-snapshot
                         (ghostel--line-mode-restore line-snapshot))))
--- a/src/Renderer.zig
+++ b/src/Renderer.zig
@@ -230,7 +230,7 @@ fn updateFontInfo(self: *Self, env: emacs.Env) bool {
     if (env.isNil(new_font)) {
         self.font_info = null;
     } else {
-        const default_font_info = env.f("query-font", .{new_font});
+        const default_font_info = env.f("ghostel--query-font-cached", .{new_font});
         // The value is a vector:
         // [ NAME FILENAME PIXEL-SIZE SIZE ASCENT DESCENT SPACE-WIDTH AVERAGE-WIDTH
         //   CAPABILITY ]
@@ -723,7 +723,7 @@ fn adjustGlyph(
     //       Most chars are covered by SOME font on the system.
     if (env.isNil(font)) return;
 
-    const font_info = env.f("query-font", .{font});
+    const font_info = env.f("ghostel--query-font-cached", .{font});
     const ascent = env.extractInteger(env.vecGet(font_info, 4));
     const descent = env.extractInteger(env.vecGet(font_info, 5));
     const height = ascent + descent;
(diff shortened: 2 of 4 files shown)


emil-e:
Thanks for the PR. This is a good improvement and I think we should merge it.

Long term, I think it's even better to cache the scaling per glyph but I can take it upon me to implement that, this at least improves the immediate case.

emil-e:
LGTM. @dakra Any concerns?

dakra:
lgtm. I'll simplify the `gc-cons-threshold` code a bit and then merge.
Don't think we need a defcustom for it and can just use the let* binding we already have for the whole redraw function.
The Emacs native `pixel-scroll.el` does also just
```
          ;; The animations are smoother if the GC threshold is
          ;; reduced for the duration of the animation.
          (gc-cons-threshold (min most-positive-fixnum
                                  (* gc-cons-threshold 3)))
```
to temproarily disable garbage collection. So think the same is fine for us and we don't need 2 more defcustoms.

dakra:
Merged. Thanks ❤️

PR #299, opened 2026-05-20
Route OSC actions through a custom stream handler

dakra:
## Summary

- Replace ~395 lines of pre-write / post-write byte scanning in `module.zig` with a custom `gt.Stream` handler (`src/GhostelHandler.zig`) that wraps libghostty's standard terminal handler and overrides the OSC arms.
- Single source of truth for OSC parsing — ghostty's parser (chunk-boundary safe, BEL/ST terminator handling, ConEmu/iTerm disambiguation, color-request lists) instead of a parallel scanner.
- OSC 51 is the only OSC still scanned by hand because it is ghostel's elisp-eval extension, not a standard OSC.

## Behavioural notes (carried in test fixtures)

- OSC 7 with an embedded ESC truncates at the second `\e]` (ghostty parser behavior); test expects `PWD="PARTIAL"`.
- Empty notifications (`\e]9;\e\\` / `\e]777;notify;;\e\\`) drop at the FFI boundary.
- OSC 9;4;1 → `progress=0` (parser pre-fill); OSC 9;4;1;`<garbage>` → `progress=nil` (parser writes null over the pre-fill). Documented quirk, no defensive default.
- ConEmu OSC 9 sub-codes (`9;5`, `9;12<anything>`) are swallowed by ghostty's parser before they can re-route to iTerm notifications.
- OSC 133 `N` (new_command) maps to the same `A` marker ghostel already uses for fresh-line prompts.

## Test plan

- [x] `make -j8 all test-evil` — 78 ert + native + evil tests, lint, byte-compile clean
- [ ] Smoke-test in `emacsclient`: prompt navigation (OSC 133), shell `pwd` tracking (OSC 7 / 9;9), color-scheme query reply (OSC 10 / 11 / 4), clipboard set (OSC 52), notification (OSC 9 / 777), progress (OSC 9;4).

### DIFF
--- a/src/GhostelHandler.zig
+++ b/src/GhostelHandler.zig
@@ -1,0 +1,259 @@
+//! Custom stream handler that delegates almost everything to libghostty's
+//! standard terminal handler but intercepts OSC-related actions so we can route
+//! them to Elisp callbacks instead of re-parsing the same bytes ourselves.
+
+const std = @import("std");
+
+const emacs = @import("emacs.zig");
+const gt = @import("ghostty-vt");
+const GhostelTerm = @import("terminal.zig");
+
+const Self = @This();
+
+inner: gt.TerminalStream.Handler,
+
+pub fn init(terminal: *gt.Terminal) Self {
+    return .{ .inner = .init(terminal) };
+}
+
+/// Called by `gt.Stream.deinit`.
+pub fn deinit(self: *Self) void {
+    self.inner.deinit();
+}
+
+/// Dispatcher invoked by `gt.Stream` for every parser action.  Anything
+/// outside the OSC arms below is forwarded verbatim to the standard
+/// handler so terminal state stays consistent.
+pub fn vt(
+    self: *Self,
+    comptime action: gt.StreamAction.Tag,
+    value: gt.StreamAction.Value(action),
+) void {
+    switch (action) {
+        // For `semantic_prompt` and `color_operation` we forward FIRST
+        // so the standard handler updates terminal state (per-row
+        // semantic flag, palette/dynamic color sets), then fire the
+        // Elisp callback / emit the query reply.
+        .semantic_prompt => {
+            self.inner.vt(action, value);
+            self.handleSemanticPrompt(value);
+        },
+        .color_operation => {
+            self.inner.vt(action, value);
+            self.handleColorOperation(value);
+        },
+
+        // For these, the standard handler is a no-op (see
+        // `stream_terminal.zig` — they are listed in the "no
+        // terminal-modifying effect" arm), so we handle them
+        // entirely here.
+        .report_pwd => self.handleReportPwd(value),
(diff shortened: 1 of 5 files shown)


emil-e:
Huge improvement!

src/GhostelHandler.zig
```diff
@@ -0,0 +1,271 @@
+//! Custom stream handler that delegates almost everything to libghostty's
+//! standard terminal handler but intercepts OSC-related actions so we can route
+//! them to Elisp callbacks instead of re-parsing the same bytes ourselves.
+//!
+//! Why a custom handler instead of byte-scanning:
```
emil-e: I think these types of comments can go away. Nobody seeing this code would think "hey... why don't they duplicate all this code manually?"

src/GhostelHandler.zig
```diff
@@ -0,0 +1,271 @@
...
+//!
+//! libghostty's `Parser` already implements the full OSC state machine -
+//! chunk-boundary safe, BEL/ST terminator handling, ConEmu vs iTerm
+//! disambiguation, color-request lists.  By wrapping the standard handler
+//! and overriding the OSC action arms, we get the parser for free and
+//! stop maintaining a parallel scanner.
+//!
+//! Only OSC 51 (ghostel's elisp-eval extension) still needs a bespoke
```
emil-e: Document that at the point of the OSC51 scanner instead.

src/module.zig
```diff
@@ -297,406 +293,66 @@ fn fnWriteInput(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value, _: ?*
...
-                continue;
+/// Single-pass scan: the `intermediate` introducer carries the literal
+/// "E" so we can match the full prefix in one `indexOfPos`.  Limitation:
+/// an OSC 51 split across two `ghostel--write-input` calls is dropped
+/// entirely — we keep no carry-over state between calls.  Unlike
+/// ghostty's `Parser`, which buffers an OSC body across chunks, this
+/// scanner only sees one `data` slice at a time.  Adding carry-over
+/// would require per-`GhostelTerm` state; not worth it until an
+/// OSC 51 chunk-spanning case actually shows up.
+fn dispatchOsc51(env: emacs.Env, data: []const u8) void {
```
emil-e, line 329: Doesn't libghostty have a parser for general OSCs? Might not be able to use the handler but I think we can at least us a parser?

emil-e (changes requested):

src/module.zig
dakra: I added a TODO, to replace it.
We can't use Ghostty's parser (see TODO comment).
But either they accept a change, OSC 51 is acually defined in xterm and reservered for "Emacs shell", so that would be best, but we could also use a unused clipboard sub-protocol.
I don't think someone cares if they emit `51;E` or `52:E`. but since 51 is the "Emacs shell" OSC, would be nice to use it.

PR #301, opened 2026-05-20
Add ghostel-glyph-scale-floor: configurable minimum glyph scale

### LINKED ISSUE #298: Allow CJK / wide-character glyphs to render at natural size (configurable scale floor)
## Summary

Today `adjustGlyphs` in `src/Renderer.zig` scales any glyph whose font metrics don't match `default_font_info` so it fits exactly inside its cell. For CJK characters on macOS this consistently shrinks them to ~84% of ASCII height, which looks visually unbalanced in Chinese / Japanese / Korean text.

`face-font-rescale-alist` cannot work around this, because Ghostel reads the *rescaled* glyph height and produces a reciprocal `display height` that cancels the rescale out.

## Why a knob is preferable to a behavior change

I understand the current behavior is the right default — it guarantees strict grid alignment, which is one of the main reasons to use Ghostel over Vterm. The maintainer's rationale in [#281](https://github.com/dakra/ghostel/issues/281#issuecomment-4471249410) makes the design intent very clear, and I'm not asking to change it.

But for buffers that aren't running a TUI app (a plain shell, or a session that's mostly CJK prose / output), users may want their CJK to render at the font's natural size — at the cost of CJK rows being slightly taller than pure-ASCII rows. That tradeoff is worth exposing as a config.

## Proposed configuration

Either of these would work; option 1 is more flexible.

1. **`ghostel-cjk-scale-floor`** (defcustom, float, default `0.0`)
   Clamp the computed scale: `scale = max(scale_floor, computed_scale)`. A value of `1.0` disables shrinking entirely; `0.0` keeps current behavior. One `@max` after the existing `@min` in `adjustGlyphs`.

2. **`ghostel-disable-glyph-adjust`** (defcustom, bool, default `nil`)
   Skip `adjustGlyphs` entirely when non-nil. Simpler but coarser — also disables the desirable behavior for too-large fallback Nerd-Font glyphs that #281 fixed.

I'd vote for option 1: it preserves the alignment-correction story for fallback glyphs that are way too big, while still letting CJK render at natural size when the user opts in.

## Reproduction

- Doom Emacs on macOS, ghostel module pulled with `ghostel-module-auto-install 'download`
- `doom-font` = `CaskaydiaCove Nerd Font Mono` 14pt
- Global CJK fontset = `Sarasa Mono SC` (`set-fontset-font t 'han ...`)
- Open ghostel, type any CJK character, `M-x describe-char`
- Observe: `display ((min-width (2)) (height 0.8421052631578947))`
- Setting `(setq face-font-rescale-alist '(("Sarasa Mono SC" . 1.3)))` has no visible effect — Ghostel reads the rescaled height and emits a reciprocal scale.

## Workaround

The only workaround today is to fork and patch `src/Renderer.zig` (clamp `scale` to `1.0`, or drop the height half of the display spec) and rebuild `ghostel-module.dylib`. Not great for staying current with upstream.

emil-e:
Closes #298.

## Summary

Adds a buffer-local `defcustom ghostel-glyph-scale-floor` (float, 0.0–1.0, default `0.0`) that clamps the computed glyph scale from below in `adjustGlyph`.

- At `0.0` (default) the existing strict-grid behavior is completely unchanged.
- At `1.0` CJK and other fallback glyphs render at natural font size, at the cost of slightly taller rows when fallback metrics exceed the primary cell size.
- Values between 0 and 1 allow partial relief — e.g. `0.9` prevents shrinking below 90% while still correcting grossly oversized glyphs.

Being buffer-local means different ghostel buffers (a TUI session vs. a plain shell) can use different settings independently.

## Implementation notes

- One `@max` in `adjustGlyph` (Renderer.zig) after the existing `@min`
- `FontInfo` carries `glyph_scale_floor` so a floor change triggers a full viewport invalidation without an extra `symbol-value` call per cell
- `emacs.zig` gains `asFloat(val, default)` which uses `(numberp)` + `(float)` to safely coerce integers to `f64`, avoiding a `wrong-type-argument` signal that would corrupt the env's non-local exit state and silently break all subsequent API calls in the same redraw; `std.math.clamp` enforces the `[0.0, 1.0]` range on the Zig side regardless of what is set programmatically
- ERT test: `ghostel-test-glyph-scale-floor-clamps-scale` feeds a glyph larger than the cell with floor `1.0` and asserts it is not shrunk

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -278,6 +278,15 @@ or backing scale factor."
   :type '(choice (const :tag "Auto-detect from display DPI" auto)
                  (number :tag "Explicit ratio")))
 
+(defcustom ghostel-glyph-scale-floor 0.0
+  "Minimum scale for glyphs whose font metrics don't fit the cell.
+0.0 (default) preserves strict grid alignment.  1.0 disables
+shrinking entirely so CJK and other fallback glyphs render at
+natural size, potentially making rows slightly taller and cells slightly wider."
+  :type '(float 0.0 1.0)
+  :local t
+  :group 'ghostel)
+
 (defcustom ghostel-kitty-graphics-storage-limit (* 320 1024 1024)  ; 320 MiB
   "Kitty graphics image storage cap, in bytes, per terminal.
 
@@ -4185,8 +4194,8 @@ mis-rendering is visible as an error instead of silent."
             (list src-x src-y src-w src-h pixel-w pixel-h))))
 
 (defun ghostel--kitty-apply-row-slice (row cw ch img
-                                            vp-col-clamped visible-cols
-                                            slice-x slice-w)
+                                           vp-col-clamped visible-cols
+                                           slice-x slice-w)
   "Apply one row of the sliced image at point.
 ROW is the slice index (0-based) into the image's grid of cells.
 CW / CH are the cell pixel dimensions.  IMG is the Emacs image object.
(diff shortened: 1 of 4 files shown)


emil-e:
@claude Review this

emil-e:
Reopening from fork (emil-e/ghostel) instead of pushing directly to origin.

PR #306, opened 2026-05-21
Skip CRLF normalization on alt screen

devsunb:
## Summary

Skip bare-LF -> CRLF normalization while the terminal is on the alternate screen.

## Why

Emacs PTYs lack `ONLCR`, so the module inserts `\r` before each bare `\n` - correct for line-oriented output on the primary screen. But alt-screen apps (tmux, vim, less) emit VT-correct LF (cursor down, **column preserved**); normalizing those to `\r\n` corrupts their layout.

## Fix

Skip the conversion when the active screen is the alternate one:

```zig
if (term.terminal.screens.active_key == .alternate) {
    term.vtWrite(raw);
    ...
}
```

Detection via `screens.active_key` covers all three alt-screen entry modes (DECSET 47 / 1047 / 1049) uniformly. `last_input_was_cr` is still maintained so a CRLF pair split across writes isn't mis-normalized into `\r\r\n`.

### Before

https://github.com/user-attachments/assets/d258aaf9-c7ba-4d29-81bf-b56976f4ba44

### After

https://github.com/user-attachments/assets/4fd2f858-0f23-47a5-b4e5-ddceb57b49e4

### DIFF
--- a/src/module.zig
+++ b/src/module.zig
@@ -327,6 +327,11 @@ fn fnWriteInput(raw_env: ?*c.emacs_env, _: isize, args: [*c]c.emacs_value, _: ?*
     // chunking (that's how the process filter already works), so no
     // scratch buffer, no allocation, no truncation fallback.
     //
+    // Skip normalization on the alternate screen.  Apps that use the
+    // alt screen (tmux, vim, less) send VT-correct sequences where bare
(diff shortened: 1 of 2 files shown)


dakra:
@emil-e you have a quick look as well but lgtm

PR #307, opened 2026-05-21
Suppress cursor when cell face background matches cursor color

devsunb:
## Summary

Hide the Emacs cursor when its color collides with the cursor cell's face background, so terminal apps that draw their own reverse-video cursor (Claude Code, Gemini CLI, etc.) stay visible.

## Why

Emacs's GUI ports (X11, mac-port, w32, pgtk) share a `set_cursor_gc` distinct-swap fallback that swaps the cursor's fg/bg with the cell's fg/bg whenever they match (`xterm.c:8017` and the mac-port/w32/pgtk equivalents). It's meant to keep the cursor visible against a matching cell, but it backfires for apps that draw their input cursor via SGR reverse-video: on a reverse-video cell the swap reproduces the colors of the surrounding non-inverse cells, so the Emacs box cursor becomes visually indistinguishable from normal text.

## Fix

Detect the collision and force `cursor-type` off so the cell's own reverse-video face represents the cursor:

```elisp
(defun ghostel--cursor-cell-conflicts-p ()
  (when-let* ((pos ghostel--cursor-char-pos)
              ((< pos (point-max)))
              (cell-bg (ghostel--cell-face-background pos))
              (cursor-color (frame-parameter nil 'cursor-color))
              (cell-rgb (color-values cell-bg))
              (cursor-rgb (color-values cursor-color)))
    (equal cell-rgb cursor-rgb)))
```

Suppression runs from a buffer-local `pre-redisplay-functions` hook rather than from `ghostel--set-cursor-style`. A one-time assignment there would not stick, since `cursor-type` is restored to its default between module render passes; the hook reapplies the suppression just before each redisplay. It also keeps the cursor off when the terminal sends DECTCEM hide (`\e[?25l`), tracked in `ghostel--cursor-visible`.

### Before

before

before2

### After

after

after2

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1523,6 +1523,13 @@ Matches Ghostty 1.2.0's `bold-color' configuration."
 (defvar-local ghostel--cursor-char-pos nil
   "The position of the terminal cursor in the buffer.")
 
+(defvar-local ghostel--cursor-visible nil
+  "Last VISIBLE value passed to `ghostel--set-cursor-style'.
+Mirrors the terminal's DECTCEM (show/hide cursor) state.  Consumed
+by `ghostel--maybe-suppress-cursor' so the redisplay-time hook can
+keep `cursor-type' off when the terminal has hidden the cursor,
+even if something else has reset `cursor-type' in the meantime.")
+
 (defvar-local ghostel--rendered-font nil
   "The font last used for rendering. Internally used by native code.")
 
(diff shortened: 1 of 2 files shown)


dakra:
@devsunb Thanks. Can you open an issue with more info and exact steps how to reproduce it?

Neither claude nor gemini have any problems here. I set my focused cursor color exactly the same as the default fg, but everything works as expected.

Also I'm not sure if a pre-redisplay-function that always runs and **hides** the cursor is the best solution.
So I would prefer if I can first reproduce the issue.

devsunb:
I found the root cause. The cursor was being redrawn by evil-mode, not by anything in this package. Sorry for the noise.

For future reference, I had been disabling evil in ghostel buffers like this:
```elisp
(with-eval-after-load 'evil
  (evil-set-initial-state 'ghostel-mode 'emacs))
```

But emacs state still keeps evil active, so the cursor was being redrawn. The fix was to exclude the buffer from evil entirely:
```elisp
(with-eval-after-load 'evil
  (add-to-list 'evil-buffer-regexps '("\\`\\*ghostel" . nil)))
```

Thanks for the patient review and for creating this package.

PR #311, opened 2026-05-22
Fix O(n²) performance in ghostel--filter-soft-wraps

plux:
## Summary

- `ghostel--filter-soft-wraps` concatenated one character at a time onto a result string, causing O(n²) allocations. This caused multi-second freezes when copying text with `M-w` in copy mode.
- Replaced with a chunk-collection approach: gather substrings between soft-wrap newlines into a list, join once at the end. Now O(n).

## Test plan

- Existing `ghostel-test-filter-soft-wraps` passes
- Manually verified: select large region in copy mode, `M-w` is now instant


emil-e:
Amazing, thanks!

PR #312, opened 2026-05-22
Add ghostel-comint: libghostty VT parser as a comint preoutput filter

dakra:
## Summary

New per-buffer minor mode `ghostel-comint-mode` (and global counterpart `ghostel-comint-global-mode`) that replaces `ansi-color-process-output` with a full libghostty parser running as a `comint-preoutput-filter-functions` entry. Targets the same niche as `xterm-color` but without an additional dependency — if you have ghostel installed, you already have the VT engine.

What this gets you over stock ansi-color:

- Full SGR semantics from libghostty: 8 / 16 / 256 / truecolor, italic, faint, strikethrough, overline, inverse, invisible, curly / dotted / dashed / double underline (including the colon-separator `\e[4:3m` forms), and underline color (`\e[58…m`).
- OSC 8 hyperlinks render as clickable links — reuses `ghostel-link-map` so clicks dispatch by URI scheme (`file://` opens in Emacs, `fileref:` jumps to file:line:col, http(s) goes through `browse-url`).
- OSC 7 (working-directory report) updates `default-directory` via `ghostel--update-directory`, so the same TRAMP / hostname resolution as `ghostel-mode` applies.
- DCS / APC / SS3 sequences are consumed instead of leaking bytes into the buffer.

What it does **not** do (by design):

- CSI cursor / erase / mode-set sequences are dropped. Comint is a line-append buffer; programs like `htop` or `less` still need a real ghostel terminal (`M-x ghostel`).
- Doesn't replace `comint-carriage-motion`. CR / BS / TAB pass through unchanged.

### Architecture

Two-layer split, matching the existing pattern in this repo:

- `src/comint_filter.zig` — pure Zig type wrapping `gt.Stream(Handler)`. Handler holds cumulative SGR state, palette, default colours, link state, and accumulates style runs into an output buffer during each `feed()`. SGR state survives across calls so a `\e[31m` chunk followed by a `red\n` chunk styles `red` correctly.
- `src/module.zig` — five thin `callconv(.c)` wrappers (`fnComintMakeState`, `fnComintFilter`, `fnComintSetPalette`, `fnComintSetDefaultColors`, `comintFinalize`) that bridge the Emacs module ABI; no business logic.
- `src/style_face.zig` — shared style→face plist construction extracted from `Renderer.zig` so the renderer and the stream filter stay in sync.

### font-lock interaction

When `font-lock-mode` is on in the buffer, the elisp wrapper swaps `face` text properties for `font-lock-face` on the filter's output string. font-lock's unfontify pass strips `face` but leaves `font-lock-face` alone, matching what xterm-color and current Emacs ansi-color do. This is a *partial* shield — where font-lock's own keywords match (e.g. shell-mode's `[+-]flag` comment pattern), `face` wins for that slice. Disabling font-lock in the buffer is still recommended for full coverage and documented in the mode's docstring.

## Test plan

- [x] `make -j8 all` — clean (37 ert test summaries, 0 unexpected)
- [x] 15 new native tests in `test/ghostel-comint-test.el` covering SGR (basic/256/truecolor/curly underline/italic+strike), state carryover across chunks, CSI/DCS stripping, CR/BS/TAB passthrough, OSC 8 hyperlinks, OSC 7 dirtrack, multibyte glyphs, and the font-lock-face swap behavior
- [x] Live smoke: `M-x shell`, `ghostel-comint-mode`, `bat --color=always Makefile` keeps bat's colors instead of going grey (this is what regressed in the original report and motivated the font-lock-face swap)
- [x] Live smoke: OSC 8 — print `\e]8;;https://example.com\e\\link\e]8;;\e\\` from a shell command, click on "link", confirm browse-url opens it
- [x] Live smoke: OSC 7 — change directories in shell, confirm `M-x pwd` reflects it without `shell-dirtrack-mode`


dakra:
@emil-e want to have a look at the zig code. zig best practices, docstrings etc.

This basically just adds a vt-parser. string in - string out. no terminal.

(And very minor, but it adds a overline prop that we forgot)

With this users don't have to use xterm-color anymore (or ansi-color, which is even less featureful).
And in my very unscientific performance test of
`time bat --paging=never README.md` in this repo under `M-x shell`,
it's twice as fast with `ghostel-comint-mode` enabled than without.

PR #319, opened 2026-05-24
Migrate elisp-eval from OSC 51;E to OSC 52;e

dakra:
## Summary

- **Migrate the elisp-eval extension from `OSC 51;E` to `OSC 52;e`.** Libghostty parses `OSC 51` into an action the standard handler ignores, so the old code reached its callback via a post-write byte scanner (`dispatchOsc51` in `GhostelTerm.zig`). `OSC 52` with a reserved kind byte (`e`) routes through libghostty's parser natively and gains cross-chunk reassembly for free. Dispatch lives in `GhostelHandler.handleClipboardContents` as a switch on `v.kind`.
- **Fix the sync-flush race when the OSC 52;e introducer ends a filter chunk.** When `\e]52;e;` lands alone at end of chunk 1 with the body in chunk 2, the regex matches chunk 1, flush empties pending, and chunk 2 has no `52;e;` to re-match on its own — without intervention the dispatch defers to the redraw timer (~33 ms) and `b4 prep --edit-cover` can race its temp-dir cleanup. New buffer-local `ghostel--osc52-eval-in-flight` carries the signal forward until the callback fires (the callback clears it as its first form).
- **Update README and CHANGELOG** to reflect the new wire format and correct an inaccurate "drops to invalid" claim.

## Breaking change

`ghostel_cmd` (bundled shell helper) now emits `\e]52;e;…\e\\`. Scripts hand-rolling the legacy `\e]51;E…\e\\` will silently no-op — migrate to the new format or call `ghostel_cmd`.

## Test plan

- [ ] `make -j8 all` green (78/78 evil, 21/21 OSC, all native + elisp targets pass)
- [ ] New OSC tests: `osc52-eval-native`, `osc52-kind-dispatch`, `osc52-eval-cross-chunk`, `osc52-mixed-kinds-one-write`, `osc52-eval-filter-flush-after-introducer`, `osc52-eval-filter-flush-mid-body`
- [ ] `ghostel_cmd find-file /tmp/foo` opens the file in bash, zsh, fish
- [ ] `b4 prep --edit-cover` opens the cover-letter file with no race
- [ ] OSC 4/10/11 color queries still sync-flush (existing `ghostel-test-osc-color-query-filter-flush` confirms)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- **Breaking:** `ghostel_cmd` (shell helper) now emits `OSC 52;e`
+  instead of `OSC 51;E`.  Libghostty parses OSC 51 into an action
+  the standard handler ignores, so the elisp-eval extension never
+  reached ghostel's callback through normal action dispatch — the
+  old code worked around this with a post-write byte scanner.  The
+  extension now rides on OSC 52 with a reserved `kind` byte (`e`),
+  which libghostty dispatches directly to ghostel's handler.
+  Scripts that hand-roll the legacy `\e]51;E…\e\\` sequence will
+  silently no-op; update to `\e]52;e;…\e\\` or call the bundled
+  `ghostel_cmd` helper.  The post-write byte scanner is gone; the
+  OSC 52 path is parsed by libghostty and now reassembles payloads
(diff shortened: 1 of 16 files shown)


dakra:
@emil-e code changes are minimal and our own osc51 zig parser is removed. wdyt?

emil-e:
I say go for it. It would be nice if we could somehow get rid of the flushing business here, but that can be a separate thing for the future.

PR #326, opened 2026-05-25
Centralize ghostel buffer terminal initialization

emil-e:
## Summary

- Merges `ghostel--prepare-buffer` into a redesigned `ghostel--init-buffer` that is the single invariant boundary between an Emacs buffer and its native terminal handle: it clears stale state, creates the terminal, and attaches it as a buffer-local immediately
- Adds `ghostel--create` as a convenience wrapper that creates a fresh buffer, calls `ghostel--init-buffer`, and returns the buffer — used by both production code and tests
- Updates tests to use `ghostel-test--with-terminal-buffer` (and the evil variant `evil-ghostel-test--with-buffer`) so they go through the same initialization path as production, eliminating per-test boilerplate and the need to manually set `ghostel--term-rows`

## Test plan

- [ ] Run the full test suite: `make test`
- [ ] Verify native-tagged tests pass with the module loaded
- [ ] Check that `ghostel--create` buffers are properly cleaned up (no leaked buffers after tests)


emil-e:
@dakra There's none of the breaking stuff that I mentioned before in this PR, by the way. This is just a refactoring to improve consistency of usage when it comes to internal functions.

PR #345, opened 2026-05-28
evil-ghostel: keep Ctrl passthrough active in alt-screen

0WD0:
Insert-state Ctrl passthrough used `evil-ghostel--active-p`, which intentionally returns nil in DECSET 1049 alt-screen mode. This caused `C-u` to fall back to Evil's insert-state binding and delete buffer text instead of being sent to the terminal.

Introduce a narrower predicate for Ctrl passthrough that remains active in semi-char mode during alt-screen sessions, leaving the broader Evil editing/cursor-sync predicate unchanged.

Add a regression test covering `C-u` passthrough in alt-screen mode.


dakra:
Merged, thanks.

PR #351, opened 2026-05-30
Match Ghostty's word boundaries for double-click selection

### LINKED ISSUE #350: Match Ghostty's word boundaries for double-click selection
## Request

Make `ghostel-mode`'s syntax table treat path/URL/hostname punctuation as word constituents, matching Ghostty's `selection-word-chars` default.

## Why it matters

Since #339, double-click in semi-char mode promotes via `mouse-set-point` → `bounds-of-thing-at-point 'word`, which reads the buffer's syntax table. `ghostel-mode` inherits `fundamental-mode`'s table, so `.`, `/`, `~`, `-`, `@` are word boundaries.

In a real shell prompt that gives:

- `api.example.com` — double-click selects only `example`
- `~/src/foo/bar.txt` — only `src` / `foo` / `bar`
- `git@github.com` — only `git`

None of that is what users want to copy from a terminal, and it's not what Ghostty does.

## Proposal

Define `ghostel-mode-syntax-table` (picked up automatically by `define-derived-mode`) built from a new `ghostel-word-boundary-string` defcustom, defaulting to Ghostty's set from [`src/terminal/selection_codepoints.zig`](https://github.com/ghostty-org/ghostty/blob/main/src/terminal/selection_codepoints.zig):

```
SPC TAB ' " │ ` | : ; , ( ) [ ] { } < > $
```

Everything else printable becomes word syntax. The `:set` handler rebuilds the table in place so live buffers pick up the change.

Side benefit: Evil `iw`/`aw`, `M-f`/`M-b`, and isearch word matching also become path/URL-aware in ghostel buffers.

Happy to PR this with a test if the API shape (string vs. codepoint list, name) looks right.

bittopaz:
Closes #350.

## Summary
- Add `ghostel-word-boundary-string`, a defcustom defaulting to Ghostty's `selection-word-chars` set — SPC TAB `'` `"` U+2502 `` ` `` `|` `:` `;` `,` `(` `)` `[` `]` `{` `}` `<` `>` `$` plus NUL — sourced from [`src/terminal/selection_codepoints.zig`](https://github.com/ghostty-org/ghostty/blob/main/src/terminal/selection_codepoints.zig).
- Define `ghostel-mode-syntax-table` from that string at load time. `define-derived-mode` picks it up automatically, so double-click word selection, word motion, and word-based search use terminal-friendly boundaries.
- The `:set` handler mutates the existing table object via `modify-syntax-entry`, so live ghostel buffers — which store a reference via `set-syntax-table` — stay in sync without re-entering the mode. Same rebuild-in-place pattern as `ghostel-keymap-exceptions`.

User-visible result: a double-click inside `api.example.com` selects the whole hostname, inside `~/src/foo/bar.txt` selects the whole path, inside `git@github.com` selects the whole identifier. Evil `iw` / `aw`, `M-f` / `M-b`, and isearch word matching pick up the same boundaries.

## Notes
- Demoting `()`, `[]`, `{}`, `"`, `'` from their open/close-paren and string-quote syntax classes to plain `.` is intentional and matches Ghostty. `show-paren-mode`, `forward-sexp`, and `M-x check-parens` behave differently in ghostel buffers as a result; for terminal content that's the right trade-off.
- Ghostel only owns the printable ASCII range when rebuilding. Adding a non-ASCII codepoint to the boundary string demotes it as expected; removing one previously added does not undo the explicit syntax-table entry without restarting Emacs. Documented in the defcustom docstring.

## Verification
- `make byte-compile` — clean with `byte-compile-error-on-warn t`.
- `make checkdoc`, `make docquotes`, `make package-lint` — clean.
- `make test` — all elisp test files pass, including the new `test/ghostel-word-boundary-test.el` (5 ERT tests covering hostname / path selection, boundary still splits, `mouse-start-end` mode 1 against the actual production call chain, and rebuild-on-customize).

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -794,6 +794,44 @@ is not in semi-char-mode when the drag completes."
                  (const :tag "Emacs mode"          emacs)
                  (const :tag "Do not switch"       nil)))
 
+(defcustom ghostel-word-boundary-string " \t\"'`|:;,()[]{}<>$│"
+  "Characters that terminate words in ghostel buffers.
+
+Mirrors Ghostty's `selection-word-chars' default.  Characters not listed
+here are word constituents, so double-click selects whole hostnames
+\(`api.example.com') and paths (`~/src/foo/bar.txt').
+
+The value is realized into `ghostel-mode-syntax-table', which drives mouse
+word selection and word-based motion/search.
+Use `setopt' or `customize-set-variable' so the table is rebuilt."
+  :type 'string
+  :initialize #'custom-initialize-default
+  :set (lambda (sym newval)
+         (set-default sym newval)
+         (when (boundp 'ghostel-mode-syntax-table)
+           (ghostel--rebuild-mode-syntax-table))))
+
+(defvar ghostel-mode-syntax-table (make-syntax-table)
+  "Syntax table for `ghostel-mode'.")
+
+(defun ghostel--rebuild-mode-syntax-table ()
+  "Realize `ghostel-word-boundary-string' into `ghostel-mode-syntax-table'.
+
+Mutates the table in place, so live ghostel buffers keep their
+buffer-local syntax-table reference and see customization changes.
+
+Printable ASCII starts as word syntax; configured boundaries become plain
+punctuation.  Keeping boundaries as `.' (not paren/string syntax) prevents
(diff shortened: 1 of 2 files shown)


dakra:
Merged. Thanks ❤️

PR #353, opened 2026-05-31
Overhaul the scrolling/anchoring and point restoration code

emil-e:
Previously, the Elisp code had many different heuristics to try to restore various window and buffer state across native redraws. In addition, it also had quite a lot of different hacks in order to make sure that the terminal followed scrolling output when anchored at the bottom.

This PR replaces all of this machinery with the following:
- Renderer now makes sure to restore `point`, `mark` and `window-start` for all windows showing the Ghostel buffer. This works across scrollback eviction, line replacements and all other sorts of scenarios.
- When resizing horizontally, the renderer pins the the position using the libghostty tracking pin functionality to ensure that they semantically end up in the same place after resizing.

A consequence is that the renderer no longer sets the point at the terminal cursor, Elisp instead does it manually by reading `ghostel--cursor-char-pos` when it needs to.

This massively simplifies all aspects of maintaining appropriate behavior around redraws and the Elisp code only needs to concern itself with things that it explicitly wants to modify.

Anchoring is also overhauled to be much simpler and more reliable. As a bonus, it now also anchors the bottom line with vscroll to the bottom edge of the window so that resizes no longer jitter the contents.


emil-e:
@claude Review this

Issue #357, opened 2026-06-01
What's the recommended way to update ghostel buffer names when the directory changes in ghostel?


ackerleytng:
Currently, I have

```
  (defun my-ghostel-set-title (title)
    "Set buffer name to ghostel::<CWD>::<TITLE>."
    (when (or (null ghostel--managed-buffer-name)
              (equal (buffer-name) ghostel--managed-buffer-name))
      (let* ((cwd (directory-file-name (abbreviate-file-name default-directory)))
             (new-name (if (or (null title) (string= "" title))
			   (format "ghostel::%s" cwd)
			   (format "ghostel::%s::%s" cwd title))))
	(rename-buffer new-name t)
	(setq ghostel--managed-buffer-name (buffer-name)))))

  (setq ghostel-set-title-function #'my-ghostel-set-title)

  (advice-add 'ghostel--update-directory :after (lambda (&rest _) (ghostel--set-title (ghostel--get-title ghostel--term))))
```

but it seems really clunky.

dakra:
I merged #358 

`ghostel-set-title-function` is now named `ghostel-buffer-name-function` and is triggered by a directory change as well.
I think that should solve you problem.

dakra:
Closed with #358 
Feel free to reopen if there's something missing.

PR #358, opened 2026-06-01
Unify buffer naming into a single ghostel-buffer-name-function

dakra:
## What

Replace `ghostel-set-title-function` (kept as an obsolete alias) with a single `ghostel-buffer-name-function`, called on **both** an OSC 2 title change and an OSC 7 `cd`. It receives the terminal title and reads `default-directory` for the cwd, so one function can name buffers by title, by directory, or by a combination of both.

This makes the combined `dir + title` naming asked about in #357 a plain function instead of the previous workaround (the two-hook design forced it to reach into the native term and re-fire the title hook via advice):

```elisp
;; by title (default)  ->  *ghostel: TITLE*

;; by directory
(setq ghostel-buffer-name-function #'ghostel-buffer-name-by-directory)

;; combined (the #357 case)
(setq ghostel-buffer-name-function
      (lambda (title)
        (let ((cwd (directory-file-name (abbreviate-file-name default-directory))))
          (if (and title (not (string= "" title)))
              (format "ghostel::%s::%s" cwd title)
            (format "ghostel::%s" cwd)))))
```

## How

- The OSC 2 path passes the title directly; the OSC 7 path reads the live title from the native term via `ghostel--get-title`.
- Both go through the shared `ghostel--rename-managed` guard, which defers to a manual `rename-buffer`.
- `ghostel-buffer-name-default` (by title) and `ghostel-buffer-name-by-directory` (by directory) are ready-made formatters, offered as `:type` choices. The by-title default returns `nil` for an empty title, so a `cd` before any title leaves the name unchanged (no `*ghostel: nil*`).
- `ghostel-set-title-function` becomes a `define-obsolete-variable-alias` (0.32.0). `ghostel-eshell` / `ghostel-compile` disable renaming via the new variable.

## Tests

`test/ghostel-buffer-name-test.el`: 7 pure-elisp (formatters, title path, disabled, custom, nil-return, obsolete alias) + 4 native (`:tags native`: by-directory cd, manual-rename respected, the #357 combined case, and the nil-title-on-cd guard). `make -j8 all` passes (build, byte-compile-error-on-warn, checkdoc, package-lint, test + test-native).


ackerleytng:
Thank you so much!

alasdairmuckart:
Thank you!

PR #370, opened 2026-06-03
Throttle redraws during sustained output floods

dakra:
During a flood (e.g. `cat huge.log`) redraws fire back-to-back at
`ghostel-timer-delay` (~30fps).  Each redraw mutates the buffer and
triggers a full Emacs redisplay, which profiling shows is ~50% of CPU
during bulk output — the dominant cost, far above VT parsing (~0.1s for
10MB) or native materialization.

Detect a flood in `ghostel--invalidate` (successive redraws within ~1.5
frames of each other, tracked by the new buffer-local
`ghostel--last-redraw-time`) and throttle the redraw timer to
`ghostel-flood-redraw-delay` (default 0.1s).  Intermediate frames scroll
past faster than they can be read and their rows are evicted from the
scrollback regardless, so rendering fewer of them loses no information;
a redraw still fires within the delay after output pauses so the final
screen appears promptly.

Measured ~16% faster on `cat fake-10mb.log` (2.27s -> 1.90s) with
redraws cut from 51 to 13.  Set equal to `ghostel-timer-delay` to disable.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -333,6 +333,21 @@ feedback and stop the timer entirely when idle.  When nil, use the
 fixed `ghostel-timer-delay' unconditionally."
   :type 'boolean)
 
+(defcustom ghostel-flood-redraw-delay 0.1
+  "Redraw delay in seconds during sustained high-throughput output.
+When output streams continuously faster than it can be rendered (a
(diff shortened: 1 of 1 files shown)


dakra:
Think we have more work to do here but this version isn't it.
Closing for now.

PR #379, opened 2026-06-05
Fix ghostel-eval-cmds defcustom type

chasecaleb:
Small fix to a defcustom type signature that I found. I _think_ this is the only one with a similar issue.

Reproduction: `(setopt ghostel-eval-cmds ghostel-eval-cmds)` results in a `*Warning*` buffer with: 

`⛔ Warning (emacs): Value ‘(("find-file" find-file) ("find-file-other-window" find-file-other-window) ("dired" dired) ("dired-other-window" dired-other-window) ("message" message))’ does not match type (alist :key-type string :value-type function)`


dakra:
Merged. Thanks.

PR #381, opened 2026-06-06
Add ability to send telnet escape "C-]" bound to "C-c C-]"

### LINKED ISSUE #380: Unable to send ^] (telnet escape) to the terminal
`C-]` is bound to `abort-recursive-edit` in the global map.

For whatever reason I can't send this to a remote system with ghostel.

Context is I have SSH'ed to a remote host and telnet'ed to a serial console server from there.

That works fine, but I can't find a mode that lets me send `^]` through to escape the telnet session.

I've tried:
- `C-c C-q C-]`
- `C-c M-D` follwed by `C-]`

I've created a patch to send this using `ghostel--send-encoded`

alasdairmuckart:
Even in char mode "C-]" calls 'abort-recursive-edit' so there's no way to send a telnet escape to a session inside ghostel.

This adds a function 'ghostel-send-telnet-escape' and binds it to "C-c C-]" in 'ghostel-mode-map'

Fixes #380

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1810,6 +1810,8 @@ Input modes (`ghostel-semi-char-mode-map', `ghostel-char-mode-map',
   ;; paste payload to the subprocess instead of letting the default
   ;; `xterm-paste' insert it into the (renderer-owned) buffer.
   "<xterm-paste>"    #'ghostel-xterm-paste
+  ;; Escape codes via C-c prefix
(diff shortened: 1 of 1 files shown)


dakra:
Thanks for the issue and PR. ❤️ 

I fixed it slightly different to not need a new interactive command and also
"C-/" and the "C-M-" versions also where missing so I added them in one go.

PR #404, opened 2026-06-12
Treat the click that refocuses a frame as a focus click

### LINKED ISSUE #403: Copy mode triggers even on focus click
In the latest version of Ghostel, #392 was merged. If I understand it correctly, the intention is to _not_ trigger copy mode when I simply click in a window (frame) to bring focus to it.

However, I see copy mode being triggered even when I'm very careful to simply do a mouse click (no dragging):

https://github.com/user-attachments/assets/029f2cb8-ca29-44db-8fc0-bcbe50cc55ca

I hope the keycast window helps to show the clicks. I clicked on the terminal to the right, and then back into the top Emacs frame — and copy mode activated.

This is on Linux with Gnome, Debian Testing.

dakra:
A left-click that gives an Emacs frame input focus - clicking back in
from another application or from another Emacs frame - entered copy
mode: the window inside the frame stays selected while the frame is in
the background, and the focus-in is processed before the press command
runs, so the press looked like a click in an already-selected window
and a4e4127 put the buffer in copy-mode on release.

`ghostel--focus-change' now records per-frame focus transitions in the
frame parameters `ghostel--frame-focused' / `ghostel--frame-refocused';
`ghostel-mouse-press-or-copy-mode' consumes the pending refocus flag and
records such a press as a focus click, likewise when the frame still
reports definitely-unfocused (press dispatched before the focus-in was
processed).  `unknown' focus state (terminals without focus reporting)
is not treated as unfocused so click->copy-mode keeps working there.

Consume-on-press means the *first* left click after the frame regains
focus is the focus click, even if it comes long after the refocus
(e.g. alt-tab, type, then click) - a deliberate tradeoff over timestamp
heuristics, which misfire when event processing is delayed between the
focus-in and the click.

Fixes #403

### DIFF
--- a/test/ghostel-mouse-paste-test.el
+++ b/test/ghostel-mouse-paste-test.el
@@ -352,6 +352,7 @@ mode or otherwise interfere."
   "Release with no tracking hands off to `mouse-set-point'."
   :tags '(native)
   (let ((fake-event `(mouse-1 (,(selected-window) 1 (10 . 5) 0)))
+        (ghostel--mouse-press-was-selected t)
         (set-point-arg nil)
         (mouse-event-called nil))
     (with-temp-buffer
@@ -373,6 +374,7 @@ without forwarding the second arg, `mouse-set-point' would just
 move point and clobber the word selection set by `mouse-drag-region'."
   :tags '(native)
   (let ((fake-event `(double-mouse-1 (,(selected-window) 1 (10 . 5) 0)))
+        (ghostel--mouse-press-was-selected t)
         (set-point-promote nil))
     (with-temp-buffer
       (setq-local ghostel--term 'fake)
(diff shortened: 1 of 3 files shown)


test/ghostel-mouse-paste-test.el
```diff
@@ -551,21 +553,113 @@ switch to, so the click just sets point and enters no mode."
...
+      (setq-local ghostel--input-mode 'semi-char)
+      (unwind-protect
+          (cl-letf (((symbol-function 'ghostel--mode-enabled)
+                     (lambda (_term _mode) nil))
+                    ((symbol-function 'mouse-drag-region) (lambda (_event) nil))
+                    ((symbol-function 'select-window) (lambda (&rest _) nil))
+                    ((symbol-function 'frame-focus-state) (lambda (&optional _f) 'unknown)))
+            (set-frame-parameter nil 'ghostel--frame-refocused nil)
+            (ghostel-mouse-press-or-copy-mode fake-event)
+            (should ghostel--mouse-press-was-selected))
```
emil-e: I think it might be better to simulate an actual click (down then up) in the tests rather than checking this flag, which is an implementation detail. That would make it less fragile.

emil-e (changes requested):

test/ghostel-mouse-paste-test.el
dakra: Good suggestion. Didn't think anyone would read the tests 🙈

emil-e (approved):

dakra:
merged in main

PR #406, opened 2026-06-14
Native PTY backend for local shells

emil-e:
## Summary

Adds a native, Zig-owned PTY path for local ghostel buffers, behind the same
Elisp input and rendering API. The goal is to move PTY output processing off
the Emacs main thread: Zig opens the PTY, spawns the child, reads output on a
background thread, and feeds bytes straight into libghostty-vt — notifying
Emacs (via an event pipe) only when a terminal callback must run or the buffer
needs a redraw. High-volume output no longer has to run through an Emacs
process filter chunk by chunk.

New `ghostel-use-native-pty` (default `t`) selects the path for local shells.

## Process model

Two PTY paths share the same input + renderer code (`ghostel--spawn-process`
dispatches):

- **Native (local, default)** — `src/PtyProcess.zig` opens the PTY and
  forks/execs; `src/NativeProcess.zig` owns the read-loop thread and writes
  Lisp event forms to an Emacs pipe process for callbacks / redraw
  invalidation; `src/GhostelTerm.zig` routes writes/resizes to either backend.
- **Emacs (remote/TRAMP, or fallback)** — `make-process` owns the PTY and
  `ghostel--filter` feeds the terminal synchronously, preserving TRAMP
  file-handler behavior.

Input is written immediately through `ghostel--write-pty` on both paths (the
old Elisp input-coalescing buffer is gone). See `ARCHITECTURE.md` for the full
description.

## Redraw path

The interactive-echo immediate-redraw decision moved out of `ghostel--filter`
into `ghostel--invalidate`, so it now applies to **both** backends (the native
path reaches redraw via the event pipe, which the old filter-only logic never
covered). The byte-size threshold (`ghostel-immediate-redraw-threshold`) was
removed — it didn't carry its weight and was meaningless for the native path,
where the reader drains the FD before notifying Emacs. The decision is now
purely "did output arrive within `ghostel-immediate-redraw-interval` of a
keystroke."

## Benchmarks

`bench/` gains a real-spawn **backend comparison** (`run-bench.sh --backends`)
and a **typing-latency comparison** (`--typing`) that drive both PTY backends
through the real pipeline. On sustained multi-MB `cat`, the native path runs
roughly 2× the Emacs path and the gap widens with size; single-keystroke
latency is a wash (the native path's event-pipe hop costs a few µs only when
Emacs is idle). The misleading `pty/*` section (a pipe + stripped filter, never
an actual PTY) was removed and `e2e/*` reframed as the fair cross-emulator
comparison.

## Tests

PTY-path tests are matrixed across both backends via a new
`ghostel-test--with-pty-matrix`, plus dedicated `ghostel-spawn-test.el` and
`ghostel-native-process-test.el`, and shared helpers for driving a real child
and reading what it received. The large diff in the existing test files is
mostly this matrixing and the removal of the input-coalescing tests.

## Verification

`make test` and `make test-native` both pass.


dakra:
@claude review

emil-e:
Review comments fixed.

PR #410, opened 2026-06-15
Detect a refocus click from the focus-in event, not a focus transition

dakra:
## Problem

A left-click that refocuses an Emacs frame (clicking back into Emacs from another app, or into another Emacs frame) is meant to be a *pure focus click*: it places point and stays in semi-char mode rather than entering copy mode. In practice it works for a while and then, after some app/frame switch, **every refocus click enters copy mode** and freezes the buffer (#403 regression).

## Root cause

Detection relied on a recorded `nil`→`t` focus *transition* plus `frame-focus-state`. Both signals depend on the window system delivering `focus-out` events — which the macOS **NS port does not do reliably**. A dropped `focus-out` leaves `frame-focus-state` stuck reporting focus *and* the recorded state stuck non-transitioning, so a later refocus click becomes byte-for-byte indistinguishable from an in-window interaction click → copy mode.

This was confirmed:
- a deterministic reproduction (the stale-`t` refocus click is identical to a genuine interaction click at the press handler);
- a live probe on the real NS port (Emacs 32): `frame-focus-state` is demonstrably reachable into a stale-`t` state (e.g. via `x-focus-frame`, frame raises), `focus-out` is not delivered on many paths, and a later real focus change does not clear a stale `t`;
- `nsterm.m`: `windowDidBecomeKey` stores `FOCUS_IN_EVENT` **unconditionally**, so the refocusing click always delivers a fresh `focus-in` regardless of stale internal state.

## Fix

Flag the refocus click from the **`focus-in` event itself** instead of a transition. New `ghostel--frame-focus-flags` runs on `after-focus-change-function` only — never on the redisplay-time window hooks that also drive `ghostel--focus-change` — and marks every frame that currently reports focus. The click that brings a frame to the foreground always delivers a `focus-in` just before the press (even when an earlier `focus-out` was dropped), so reading the live focus state at each focus event is robust where transition tracking is not.

`ghostel--focus-change` keeps only its terminal focus-event duties; the unused `ghostel--frame-focused` parameter is removed. The press handler is unchanged.

### Residual trade-off

If `after-focus-change` fires *spuriously* with the frame focused *between* two interaction clicks (e.g. a command calls `raise-frame`/`select-frame-set-input-focus` mid-session), the next interaction click would be treated as a focus click and skip copy mode. This is rare in pure terminal use and is the inverse, far less disruptive failure. The cross-app case fundamentally requires the `focus-in` signal — there is no focus-event-free discriminator.

## Testing

- `ghostel.el` byte-compiles clean with `byte-compile-error-on-warn`.
- `make test test-native test-zig` → exit 0, 0 unexpected.
- Added `ghostel-test-mouse-1-click-stale-focus-is-focus-click` (the stale-`t` regression — fails on the old code) and `ghostel-test-frame-focus-flags-event-based`; replaced the old transition test.

> Note: interactive focus behaviour can't be exercised through `emacs --batch`; the unit tests + NS source confirm the mechanism, but the true cross-app *click → focus-in → no copy mode* path is best confirmed in a live GUI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,11 @@ All notable changes to this project will be documented in this file.
 - A single left-click that gives an Emacs frame input focus — clicking back
   into Emacs from another application, or into another Emacs frame — no longer
   enters copy mode; it is treated as a pure focus click like a click in a
-  previously-unselected window (#403).
+  previously-unselected window (#403).  The refocus click is now detected from
+  the focus-in event itself rather than a tracked focus transition, so it keeps
+  working when the window system's focus state is stale — notably on macOS (NS),
+  where a dropped focus-out can leave `frame-focus-state' stuck reporting focus
+  and previously made every refocus click enter copy mode after a while.
 
 ## [0.34.0] — 2026-06-08
 
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -2659,7 +2659,7 @@ in an already-focused frame, before focusing it, for
   (let* ((win (posn-window (event-start event)))
          (frame (window-frame win))
          ;; First press since the frame gained focus, or press dispatched
-         ;; before the focus-in (nil, not `unknown'): a focus click (#403).
+         ;; before the focus-in (nil, not `unknown'): a focus click.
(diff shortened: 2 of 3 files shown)


dakra:
merged in main

PR #415, opened 2026-06-17
Fix oversized last CJK character on a line

thinkiny:
The fix the CJK last character become larger:
  
  
  The space-claiming loop had three related bugs causing the last CJK
  character on a line to render larger than mid-line counterparts:

  1. The while-loop iteration expression incremented char_width before
  the body ran, so breaking on the first iteration (aspect fits,
  non-space neighbor, or trailing space) left char_width one column
  too high.  Every break path now undoes the increment.
  2. The aspect ratio check used strict < instead of <=, so when
  the glyph aspect exactly matched the cell aspect, the loop kept
  going one more column than needed.
  3. Positions past the newline were treated as freely claimable
  (continue), letting the last glyph on a line expand into
  unlimited trailing space.  Now we break instead — trailing space
  would always make the last glyph larger than the rest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ All notable changes to this project will be documented in this file.
   independently.
 
 ### Fixed
+- ZSH shell integration now reports gethostname(2) in its OSC 7 directory update
+  (matching the bash and fish integrations and Emacs `system-name`), instead of
+  zsh's `$HOST`, which is canonicalized to the fully qualified domain name when a
+  DNS search domain is configured.  Previously the FQDN disagreed with
+  `system-name` (the short hostname), so the local shell was misclassified as
+  remote and switched on TRAMP (#417).
+- Fish shell integration now caches the hostname once at load instead of forking
+  the `hostname` command on every prompt, matching the bash and zsh integrations.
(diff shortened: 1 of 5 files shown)


emil-e:
This feels like something we should fix, but I'm not sure #3 is a good idea because it feels like you still have the issue that a space can cause the last character to be larger too. It also removes part of the idea with the feature, that a symbol at the end of a line can claim extra space.

I wonder if we should just skip the claiming for CJK or similar characters.

thinkiny:
without the patch, the last character is always larger than others

emil-e:
Yes, that much I understand. I'm saying the fix might not be the right one. I'm not saying that it's correct that the last character is always larger.

emil-e:
@thinkiny Could you tell me what font you’re using for CJK and what font you’re using as a primary font? I have some ideas.

thinkiny:
CJK: Sarasa Mono SC
Primary font:  Sarasa Mono SC

emil-e:
I will debug and do some fixes. Really appreciate the report and some of the fixes in here I will use.

emil-e:
@thinkiny See if #420 fixes your issue

thinkiny:
well done, it's fixed, 👍

PR #416, opened 2026-06-17
Drain ready PTY output bursts

ezemtsov:
## Summary
- Drain bounded bursts of ready PTY output from `ghostel--filter`.
- Add `ghostel-output-drain-limit`, defaulting to 256 and disabled at 0.
- Preserve the existing redraw scheduling path.

## Background
This came from benchmarking Ghostel as the terminal workload for an Emacs rendering-performance investigation. A focused `vtebench` scrolling run showed that rendering was not the limiting factor: Ghostel was spending wall time returning to the event loop between small PTY reads while more output was already ready. This patch keeps Ghostel's existing redraw scheduling behavior, but lets the process filter drain a bounded burst of ready output before yielding.

## Benchmark
Focused local `vtebench` `scrolling` run inside Ghostel:

| Variant | Result |
| --- | ---: |
| Before | ~10.9s |
| After | ~0.35s |

## Tests
- `make -j1 .build/tests/elisp-ghostel-keys-test.ok checkdoc docquotes EMACS=/home/ezemtsov/git/emacs/src/emacs`


dakra:
We went over similar iterations as well, but no solution was really satisfactory.
E.g. drain fast for good performance and be able to run `yes` and still stay responsive.
Started with a similar patch than yours and tried a few different variants (I think some drain-* branches are still up).

In the end @emil-e added a zig native PTY. It'll be merged very soon, check out #406
Feedback welcome.
It's **very** fast and also stays responsive.

dakra:
Ah, and of course thanks for the PR anyway.  ❤️

ezemtsov:
@dakra thanks! looking forward the PR to be merged.

I'm working on extending emacs with skia backend, and this is the biggest bottleneck so far.

dakra:
> I'm working on extending emacs with skia backend, 

Nice. There are quite a few different patches floating around currently to improve the Emacs display engine.
Latest on emacs-dev mailing list that adds gpu rendering backend. Just in case you're interested and haven't seen it yet.

> and this is the biggest bottleneck so far.

what is "this"?

dakra:
@ezemtsov it's merged now.
And just for fun, as you mentioned `vtebench`, here's benchmark vs macOS Terminal.app:
image

ezemtsov:
@dakra cool numbers, which emacs build are you using?

> what is "this"?

I'm using ghostel + vtebench as a benchmark to iterate over emacs rendering performance on wayland+gpu. And ghostel draining logic was preventing from further optimizations. Testing the current master and it's much faster, so I can continue my work. Thank you.

dakra:
> @dakra cool numbers, which emacs build are you using?

MacOS, emacs-plus, master branch.

> I'm using ghostel + vtebench as a benchmark to iterate over emacs rendering performance on wayland+gpu. And ghostel draining logic was preventing from further optimizations. Testing the current master and it's much faster, so I can continue my work. Thank you.

Ah, ok. But then you must be careful now because `vtebench` measures the PTY throuput, and Ghostel now handles that completely in Zig native code. So I'm afraid you can't really use `vtebench` anymore for rendering performance.
(You can still use it, because obviously it displays the buffer, but it's not a strict dependency like it was before)

But I'm also not sure if that's the correct way to measure anyway, can't you just create a large buffer and then measure (pure elisp) how long redisplay takes?

PR #428, opened 2026-06-20
Switch input mode when point leaves the live input point

dakra:
## What

The keyboard analog of the existing mouse/mark mode-switch: in semi-char mode, when point moves off the live terminal cursor, ghostel switches to a read-only mode (`ghostel-point-leave-input-mode`, default `copy`) so the navigated position is stable instead of the next redraw yanking point back to the prompt.

## Details

- New defcustom `ghostel-point-leave-input-mode` (`copy`/`emacs`/`nil`), a sibling of `ghostel-mouse-drag-input-mode` and `ghostel-mark-activation-input-mode`.
- New public command `ghostel-maybe-leave-input` — a no-op unless point has left the live cursor in semi-char; ignores its args so it works as a hook function or `:after` advice.
- Wired into `isearch-mode-end-hook` and (deferred) `minibuffer-exit-hook` (covers `consult-line` and other minibuffer jumps). Packages with no hook of their own (avy, flash) add `ghostel-maybe-leave-input` themselves — documented in the README.
- The deferred `minibuffer-exit` re-anchor now skips copy/Emacs buffers, so a minibuffer jump that switched modes isn't snapped back to the live cursor; semi-char/char/line still re-anchor as before. The guard lives on the auto-follow path rather than in `ghostel--anchor-window`, which intentionally anchors Emacs mode for deliberate input like paste (covered by `ghostel-test-emacs-mode-yank-scrolls-to-live-cursor`).

## Testing

- New ERT tests in `ghostel-modes-test.el`; `make -j8 all` green.
- Live-verified under elate (real shell in a ghostel buffer) across both `copy` and `emacs` targets for isearch, consult-line (deep + near-bottom), avy, and flash, plus the no-false-positive typing path.

README and CHANGELOG updated.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -777,6 +777,17 @@ Mouse selection is governed separately by `ghostel-mouse-drag-input-mode'."
                  (const :tag "Emacs mode"          emacs)
                  (const :tag "Do not switch"       nil)))
 
+(defcustom ghostel-point-leave-input-mode 'copy
+  "Input mode to switch to when point leaves the live input point in semi-char.
+Triggered by any command that moves point off the terminal cursor without a
+mouse click or region activation.  Something like `isearch', `consult-line',
+`avy', `goto-line', wheel scrolling, etc.
+
+See also `ghostel-mouse-drag-input-mode', `ghostel-mark-activation-input-mode'."
+  :type '(choice (const :tag "Copy mode (default)" copy)
+                 (const :tag "Emacs mode"          emacs)
+                 (const :tag "Do not switch"       nil)))
+
 (defcustom ghostel-word-boundary-string " \t\"'`|:;,()[]{}<>$│"
   "Characters that terminate words in ghostel buffers.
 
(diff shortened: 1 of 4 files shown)


lisp/ghostel.el
```diff
@@ -6352,10 +6383,26 @@ a Ghostel window making it lose its anchoring."
                    (dolist (entry anchored)
                      (let ((win (car entry))
                            (buffer (cdr entry)))
+                       ;; Guard lives here, not in `ghostel--anchor-window',
```
emil-e: I wonder if it maybe is still better to have this guard in ghostel--anchor-window? Then it won't get triggered in the redraw path also, and we can remove that specific guard there. We could then maybe have a `force` parameter for the places where we explicitly want to anchor, such as input.

emil-e (changes requested):

lisp/ghostel.el
dakra: fixed

PR #431, opened 2026-06-20
Mark OSC 133 output and skip prompt regex on output rows

### LINKED ISSUE #430: `C-a` in copy/emacs mode stops mid-line on command output
I hit an issue where in emacs-mode `C-a` did not go to the beginning of a line.

**What I did**:

1. In a ghostel buffer, run `dnf update`. There will be multiple lines like the following:

   ```
   [44/48] Removing docker-ce-cli-1:29.5.3-1.fc43.x86_64                          100% |  14.3 KiB/s | 190.0   B |  00m00s
   ```
2. Enter emacs mode (`C-c C-e`)
3. Put point somewhere on that line and press `C-a`

**Expected**: Point moves to the beginning of the line (column 0). 

**Actual**: Point moves to the first `|`, i.e. right after `100% ` 

While the docstring of `ghostel-prompt-regexp` already acknowledges this class of false positive ("lines like `5 > 3` will yield false positives for column-aware motions"), I think there may be a fix for the specific case where OSC 133 shell integration is active (`133;A/B/C/D` are all emitted).

In that case, I believe ghostel receives `133;C` and tags the output cells internally as `semantic_content = .output`. But the renderer only turns `.prompt` and `.input` into text properties, not `.output` ([src/Renderer.zig#L380](https://github.com/dakra/ghostel/blob/adb010b7fec943405006fcd1fac280e74ffa9e30/src/Renderer.zig#L380)):

```zig
switch (props.semantic_content) {
    .prompt => // put-text-property ... ghostel-prompt t
    .input  => // put-text-property ... ghostel-input  t
    else => {},   // .output → no property
}
```

So an output line ends up indistinguishable from a raw line with no shell integration, and the regex fallback runs on it.

Assuming it makes sense to also handle `.output` here, a fix might look something like:

1. In `src/Renderer.zig`, tag output cells too: `.output => put-text-property ... ghostel-output t`
2. In `ghostel-beginning-of-input-or-line`, skip the regex fallback when the line carries `ghostel-output`.

If that approach looks reasonable, I'm happy to submit a patch.

```markdown
### System info
System type:         gnu/linux
Emacs version:       30.2
Ghostel version:     0.35.4
```

psschwei:
Fixes #430 

Shells with OSC 133 integration emit a C marker to delimit command output. Previously these output cells carried no Emacs text property, so `ghostel-beginning-of-input-or-line` (C-a) had no way to tell output apart from a prompt and fell back to `ghostel-prompt-regexp`. On output that happens to look prompt-like (e.g. a progress bar such as "[44/48] Removing docker-ce-cli 100% | 14.3 KiB/s") the regex matched mid-line and C-a jumped past real text instead of to column 0.

Propagate the cell's `.output` semantic content to a new `ghostel-output` text property and skip the regex fallback on any row carrying it.

Because libghostty's default cell semantic is `.output`, a guard (`semantic_output_enabled`) ensures the property is only applied once an OSC 133 marker has actually been seen, otherwise every cell on a plain shell would be marked as output.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -3915,7 +3915,9 @@ pressing \\`C-a' gives the standard column-0 behaviour."
                 (and (> pos bol) (< pos eol) pos)))))
          ;; Regex fallback for shells/REPLs without OSC 133.
          (regex-target
-          (unless (or line-mode-target prop-target)
+          (unless (or line-mode-target
+                      prop-target
+                      (text-property-not-all bol eol 'ghostel-output nil))
             (ghostel--regex-prompt-end bol))))
     (cond
      (line-mode-target (goto-char line-mode-target))
--- a/src/GhostelTerm.zig
+++ b/src/GhostelTerm.zig
@@ -78,6 +78,10 @@ pub fn redraw(self: *Self, force_full: bool) !void {
     defer self.unlockTerm();
 
     const env = emacs.current_env orelse return;
+    // A native process sees OSC 133 markers on its own thread; fold its flag in.
+    if (self.process) |proc| {
+        if (proc.semantic_output_enabled) self.renderer.semantic_output_enabled = true;
+    }
     const pre_size = .{ self.terminal.cols, self.terminal.rows };
     try self.renderer.redraw(self.alloc, env, force_full);
     _ = env.f("ghostel--kitty-clear", .{});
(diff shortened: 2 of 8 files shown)


dakra:
@psschwei thanks for the PR. CI needs some fixes.

emil-e:
I'm not really in love with the `semantic_output_enabled` flag. Is it to distinguish from the case where it's all marked as output due to missing shell integration?

psschwei:
> I'm not really in love with the `semantic_output_enabled` flag. Is it to distinguish from the case where it's all marked as output due to missing shell integration?

Yes, that was my intent. My reading was that `.output` can also be the default semantic value when there is no OSC 133 shell integration, so marking every `.output` cell as `ghostel-output` might disable the regex fallback for plain shells/REPLs. The flag was meant to only treat `.output` as meaningful after we’ve actually seen OSC 133 markers.  Happy to try something else if you think that would make more sense.

psschwei:
Fixed in #464

PR #438, opened 2026-06-20
Extract line mode into ghostel-line-mode.el

dakra:
## Summary

Splits the self-contained **line mode** feature out of the ~6.9k-line `ghostel.el`
into a new companion file `lisp/ghostel-line-mode.el`, required from `ghostel.el` —
mirroring the existing `ghostel-kitty.el` split. Pure refactor, no behavior change:
`ghostel.el` shrinks by ~830 lines and the moved code is byte-for-byte identical,
only relocated.

## What moved
- The 4 `ghostel-line-mode-*` defcustoms (now with explicit `:group 'ghostel`,
  since the new file has no in-file `defgroup`).
- Line-mode state vars and the `ghostel-line-mode-map` keymap.
- All `ghostel-line-mode-*` / `ghostel--line-mode-*` commands and helpers, plus the
  optional `bash-completion` integration.

## What stayed in core (deliberately)
Shared helpers used by the core `C-a` binding, the `evil-ghostel` extension, and
tests: `ghostel--regex-prompt-end`, `ghostel-input-start-point`,
`ghostel-beginning-of-input-or-line`, and the cursor-state queries
(`ghostel-cursor-point`, `ghostel--viewport-row-at`,
`ghostel-point-on-cursor-row-p`). (`ghostel--saved-hl-line-mode` is `hl-line-mode`,
not ghostel line mode — also stayed.)

## Keymap parent
`ghostel-line-mode-map` previously used `:parent ghostel-mode-map`, a load-time
reference to a core value. To avoid coupling load order, `:parent` is dropped and the
parent is wired at runtime via `set-keymap-parent` in `ghostel.el`, right after
`ghostel-mode-map` is defined. This lets `(require 'ghostel-line-mode)` sit at the top
of `ghostel.el` with no load-order dependency, exactly like `ghostel-kitty.el`.

## Verification
- `make -j8 all` passes (build + elisp/native/zig tests + lint). The build runs with
  `byte-compile-error-on-warn t`, so any missing forward declaration would fail it.
- Full `ghostel-line-mode-test.el` suite and `evil-ghostel` (109/109) pass.
- Live-tested against a real bash shell: mode entry/editing/send/exit; history +
  `C-a` + self-insert point-snapping + `S-RET`; and `TAB` completion / `C-c C-c`
  interrupt / alt-screen auto-pause-resume — all behave identically to before.


lisp/ghostel-line-mode.el
```diff
@@ -0,0 +1,881 @@
...
+;; The mode keymap's parent (`ghostel-mode-map') is wired at runtime in
+;; ghostel.el (see the `set-keymap-parent' call there), so this file has no
+;; load-time dependency on core ghostel values and can be required early.
+
+;;; Code:
+
+(require 'comint)
+
+;; Forward declarations for core ghostel symbols (defined in ghostel.el,
+;; or the native module).  Mirrors ghostel-kitty.el.
```
emil-e: Mention of ghostel-kitty.el seems like a Claude-ism :)

emil-e (changes requested):

PR #448, opened 2026-06-22
Start remote shells as login+interactive shells

### LINKED ISSUE #444: Remote shell not started as interactive/login shell
Hi!

Thanks a lot for this lovely package.

I use ZSH on NixOS and I noticed that ghostel didn't load `.zshrc` when `default-directory` was a remote path (i.e when the shell is launched via SSH).
I had a closer look into it, and it seems that `ghostel--resolve-shell-spec` hardcodes `nil` for the shell's extra args on remote sessions, so the shell can't be called with `-i`  (interactive) or `-l (login).

I confirmed this with the following hack, which makes the remote shell load `.zshrc`.
```
(advice-add 'ghostel--resolve-shell-spec :filter-return
  (lambda (spec)
    (if (file-remote-p default-directory)
        (cons (car spec) '("-i" "-l"))
      spec)))
```

Related issue: https://github.com/dakra/ghostel/issues/285

Thanks a lot!

dakra:
## Summary

Remote (TRAMP) shells were spawned with no extra args: they were never login shells (skipping `.zprofile`/`.zlogin`), and interactivity was left to the shell's tty auto-detection, which doesn't hold on every TRAMP method (so `.zshrc` could be skipped too).

This passes `-l -i` by default for recognized shells (bash, zsh, fish) so remote sessions reliably source the user's rc/profile files, mirroring an interactive `ssh host` login. Unrecognized shells (e.g. `/bin/sh`) are left unchanged.

`ghostel-tramp-shells` entries gain an optional trailing args slot — `(METHOD SHELL [FALLBACK [ARG...]])` — to override the default per method.

Fixes #444.

## Implementation

- `ghostel--tramp-get-shell` → `ghostel--tramp-shell-spec`: now returns `(PROGRAM . EXTRA-ARGS)`, parsing the optional per-method args after the fallback slot. Program resolution (incl. `login-shell` `getent` detection) is unchanged.
- New `ghostel--default-remote-shell-args`: returns `("-l" "-i")` for bash/zsh/fish (detected by basename, so NixOS-style paths work), `nil` otherwise.
- `ghostel--resolve-shell-spec` surfaces explicit per-method args for remote sessions; the type-aware default is applied in `ghostel--start-process` only when remote shell integration is **off** — integration owns startup-file loading via its own invocation (e.g. bash `--rcfile`, which a login shell would ignore), so the default must not be appended on top of it. Local and integration-on paths are unchanged.

## Testing

- `make -j8 all` is green; 10 new ERT tests cover arg parsing, the type-aware default, and spawn composition (including a guard that integration doesn't get a stray `-l`).
- Verified live by driving a ghostel terminal to a real remote host (`/ssh:<host>:` with zsh), non-destructively:
  - **Before:** `… exec /usr/bin/zsh` → `interactive=on login=off`
  - **After:** `… exec /usr/bin/zsh -l -i` → `interactive=on login=on`

  (On that host interactivity was already auto-detected from the PTY; the observable delta there was the login flag. Passing `-i` explicitly makes interactivity reliable on TRAMP methods where the remote stdin isn't a tty — the original report's case, where `.zshrc` wasn't loading.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,15 @@ All notable changes to this project will be documented in this file.
   method.
   Fixes [#444](https://github.com/dakra/ghostel/issues/444).
 
+### Fixed
+- Remote shell integration: the per-session temp files (zsh `ZDOTDIR`, bash
+  `--rcfile`, fish init script, and the pushed terminfo dir) were deleted
+  immediately after the asynchronous spawn, racing the remote shell that had
+  not yet read them. The shell would then start with `ZDOTDIR` pointing at a
+  deleted directory, so neither the integration nor the user's own config
+  loaded. Cleanup is now deferred to buffer-kill, past the shell's startup read
+  and the session-long lifetime of the terminfo dir.
+
 ## [0.37.0] — 2026-06-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,24 @@ All notable changes to this project will be documented in this file.
   buffer that has since moved elsewhere).
   Fixes [#447](https://github.com/dakra/ghostel/issues/447).
 
+### Changed
+- Remote (TRAMP) shells now start as login+interactive shells by default for
+  recognized shells (bash, zsh, fish), so they reliably source the user's
+  rc/profile files (`.zshrc`, `.zprofile`, …) like an interactive `ssh host`
+  login. Previously remote shells were spawned with no flags: they were never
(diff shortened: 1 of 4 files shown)


edgar-vincent:
Thanks a lot for this. It seems to work as expected so far, except, by design, when remote shell integration is enabled, in which case the user's own shell configuration isn't loaded (for a trivial example, prompt colours are not set).

Could there be a way around that?

Thanks again!

dakra:
@edgar-vincent thanks for testing. Can you try again?

edgar-vincent:
This seems to work fine. Thank you so much!

Issue #429, opened 2026-06-20
⬝ (BLACK VERY SMALL SQUARE) renders as non-monospace


kalomak:
## Minimal repro:

```bash
printf '%s\n' '--------' '⬝⬝⬝⬝⬝⬝⬝⬝'
```

## Symbol

https://www.unicodepedia.com/unicode/miscellaneous-symbols-and-arrows/2b1d/black-very-small-square/
`⬝`
Name: `BLACK VERY SMALL SQUARE`
Code point: `U+2B1D`
UTF-8 bytes: `E2 AC 9D`

## Why

Can cause some visual problems like here:

https://github.com/user-attachments/assets/66fd51b2-5176-4b7e-a934-6808d2afe10d

emil-e:
Can you tell me:
- What primary monospace font you're using
- What font Emacs picks as a fallback for those characters. You can to `describe-char` with point placed on that character in in copy mode to find out.

kalomak:
Primary monospace font:
JetBrainsMono Nerd Font

Fallback for ⬝:
STIX Two Math

describe char output:
```text
             position: 57 of 229 (24%), column: 10
            character: ⬝ (displayed as ⬝) (codepoint 11037, #o25435, #x2b1d)
              charset: unicode-bmp (Unicode Basic Multilingual Plane (U+0000..U+FFFF))
code point in charset: 0x2B1D
               script: symbol
               syntax: _ 	which means: symbol
             category: .:Base
             to input: type "C-x 8 RET 2b1d" or "C-x 8 RET BLACK VERY SMALL SQUARE"
          buffer code: #xE2 #xAC #x9D
            file code: #xE2 #xAC #x9D (encoded by coding system utf-8-unix)
              display: by this font (glyph code):
    mac-ct:-*-STIX Two Math-regular-normal-normal-*-14-*-*-*-p-0-iso10646-1 (#xC75)
Character code properties: customize what to show
  name: BLACK VERY SMALL SQUARE
  general-category: So (Symbol, Other)
  decomposition: (11037) ('⬝')
```

emil-e:
Thank you, will try to reproduce!

kalomak:
This was a font issue. JetBrainsMono Nerd Font just does not have this symbol. Fix if someone has same issue is to replace the missing symbol with an existing one:

```elisp
(aset standard-display-table #x2b1d (vconcat "▪"))
```

emil-e:
[...]

PR #453, opened 2026-06-23
Scale emoji and wide glyphs with text-scale

adisbladis:
When a Ghostel buffer's font is enlarged, regular text grows but emoji & any other glyph that the renderer scales into the monospace grid stay pinned at the base font size.

- Before
repro-before

- After
repro-after

Possibly related: https://github.com/dakra/ghostel/issues/429.


emil-e:
Nice! I left this as unimplemented myself as I don't use text scaling, but I figured someone would eventually need it. I'll have a look as soon as I can.

emil-e:
@adisbladis I took the main shape of your idea and but made some simplifications, mainly how the remapped font is determined so that it doesn't require any content. You can try it out on main and se if it works for you.

PR #470, opened 2026-06-25
Fix launching Bash remotely when shell integration is enabled

keithhub:
Attempting to start ghostel for a remote Bash shell over TRAMP with `ghostel-tramp-shell-integration` set to `t` fails with:

```
=== Ghostel Debug Log ===

[14:05:49.706] RESIZE: 116x59 → 116x59 (0.1ms)
[14:05:53.763] RESIZE: 116x59 → 116x59 (0.0ms)
[14:05:55.842] RESIZE: 116x59 → 116x59 (0.1ms)
[14:05:55.901] FILTER: 7 bytes: "[H[2J"
[14:05:55.908] FILTER: 491 bytes: "/bin/bash: --: invalid option
Usage:	/bin/bash [GNU long option] [option] ...
..."
```

The failing command is roughly `/bin/bash -i --rcfile /tmp/ghostel.bash`.

Bash requires all multi-character (long) options come before any single-character options in order to be recognized. This change reorders `shell-args` to ensure that all single-character options are moved to the end. With the fix, the generated Bash command line looks like `/bin/bash --rcfile /tmp/ghostel.bash -i` and Bash launches cleanly.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -4464,6 +4464,16 @@ run the shell on the remote host."
                                (ghostel--default-remote-shell-args
                                 shell remote-integration)))
                       integration-args))
+         ;; Bash does not recognize multi-character options that come after
+         ;; single-character ones, so move single-character options to the end.
+         (shell-args (if (eq shell-type 'bash)
+                        (cl-loop for arg in shell-args
+                                 if (and (string-prefix-p "-" arg)
+                                         (not (string-prefix-p "--" arg)))
+                                 collect arg into singles
+                                 else collect arg into others
+                                 finally return (append others singles))
+                      shell-args))
          (extra-env (append
                      (unless remote-p
                        (list (format "EMACS_GHOSTEL_PATH=%s" ghostel-dir)))
--- a/test/ghostel-spawn-test.el
+++ b/test/ghostel-spawn-test.el
@@ -245,7 +245,7 @@ A login bash ignores the `--rcfile' the bash integration relies on, so
                (default-directory "/ssh:host:/home/u/"))
           (ghostel--start-process)
           (should (equal "/bin/bash" (car spawn)))
-          (should (equal '("-i" "--rcfile" "/tmp/ghostel.bash") (cdr spawn)))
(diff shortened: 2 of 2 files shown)


emil-e:
Thanks for the PR! This looks good to me but I'll let @dakra have a look, he knows more about these parts.

dakra:
@keithhub thanks for the PR / bug report.

The loop in this PR is slightly buggy.
A `ghostel-(tramp-)shell` with e.g. `("-O" "extglob")` would get reordered to `("extglob" "-O")`.

But imho we don't need the loop after all. Look at #474
I just re-ordered the way **WE** pass the options to remote bash so it works by default.
And when a user customizes `ghostel-(tramp-)shell` I just added a docstring with a "warning/info" about this bash caveat.
But otherwise it's the users responsibility that his customization work. We don't change the order for him.

wdyt?

emil-e:
@dakra That works as long as it never concatenates. But I agree that it's hard to do this correctly - you'd have to know what parameters have an arg and which ones that don't and I don't think we should reason about that.

dakra:
> That works as long as it never concatenates

Not sure I understand. what works, what concatenates?

In my PR we just fix *our* site of the problem and I could even argue that it's wrong to change a user customization for him by sorting it.

emil-e:
@dakra I meant that as long as we don't concatenate multiple soruces of bash flags, it's fine. As long as the customization just replaces whatever we do, it's fine .

dakra:
It's always appended to our 'integration args' that we only sometimes set.
and for bash that's only long-form args (`--posix`, `--rcfile`), so whatever the user customizes after that works.

Local fish and zsh have none but they don't have that restriction with long before short arg anyway.

emil-e:
@dakra Then I agree, your solution is the best one. We just have to make sure to never use short form options because then we break user customization.

keithhub:
Makes sense to me -- I originally did what #474 does, but wasn't sure if was safe to assume that `integration-args` would always be long-form options (or, alternatively, that `extra-shell-args` would always be short-form). I hadn't considered that someone may want to use `-O` though, so my solution still wasn't a general fix.

The simpler fix with a comment seem great.

Thanks for the quick response!

Issue #468, opened 2026-06-25
Add a setting to enable transparent background


ygohko:
When running Ghostel on a terminal with a transparent background setting, only Ghostel's buffer retains opaque background.
Note a attached screenshot comparing Ghostel, vterm, and scratch buffers.

I would like to request the addition of a transparent background setting.

## Reporter's environment

* ghostel version 20260618.854
* GNU Emacs 30.2
* Windows Terminal version 1.24.11321.0
* Ubuntu 26.04 on WSL2 on Windows 11 Pro 25H2

<img width="2110" height="1255" alt="Image" src="https://github.com/user-attachments/assets/ca7f810f-f854-4fc4-a8a4-dca4f9fd88d4" />

emil-e:
This is in the works right now, FYI!

emil-e:
@ygohko Can you please check if this is working now with the latest release? Unless you have explicitly set a background color on the default face, you should get a transparent background.

ygohko:
I have confirmed that latest Ghostel 20260626.1255 has transparent background.

Thank you!

<img width="2110" height="1255" alt="Image" src="https://github.com/user-attachments/assets/41a56273-78a7-49a1-b857-e4bf877ca2aa" />

PR #471, opened 2026-06-25
Use buffer face remapping for default terminal colors

emil-e:
## Summary
- remap Ghostel buffers to `ghostel-default` instead of painting terminal default fg/bg as cell face properties
- keep protocol default colors for OSC 10/11 replies while omitting default fg/bg from rendered cells
- update renderer/comint tests and docs for the new default-face behavior

## Verification
- `make build`
- `make byte-compile`
- render native + non-native tests
- comint native tests
- OSC native tests
- modes tests
- `zig build test`


emil-e:
Fixes #468

PR #502, opened 2026-07-02
Respect Zig install prefix for native module

emil-e:
## Summary
- install the native module and version sidecar under Zig's install prefix instead of escaping it
- update Makefile, CI/release builds, docs, and helper scripts to pass `--prefix .` when checkout-root artifacts are expected
- keep Emacs auto-builds safe by building into a temporary prefix and renaming the module/sidecar into place

## Verification
- `emacs --batch -Q -L lisp --eval '(setq byte-compile-error-on-warn t)' -f batch-byte-compile lisp/ghostel-module-install.el`
- `emacs --batch -Q -L lisp -L test --eval '(setq load-prefer-newer t)' -l ert -l test/ghostel-test-helpers.el -l test/ghostel-module-test.el --eval "(ert-run-tests-batch-and-exit '(not (tag native)))"`
- `zig build --prefix /private/tmp/ghostel-pr-prefix-test -Doptimize=Debug -Dcpu=baseline`
- `git diff --check`

### DIFF
--- a/test/ghostel-module-test.el
+++ b/test/ghostel-module-test.el
@@ -196,6 +196,21 @@ file mmap'd keeps a valid mapping (issue #247)."
                                 (expand-file-name "~/custom/ghostel/")))
                      (downcase (ghostel--module-directory)))))))
 
+(ert-deftest ghostel-test-module-build-dir-is-reset ()
+  "The temporary build directory is reset before use."
+  (let* ((dir (make-temp-file "ghostel-test-build-dir" t))
+         (build-dir (expand-file-name ".ghostel-build/" dir))
+         (stale (expand-file-name "stale" build-dir)))
+    (unwind-protect
+        (progn
+          (make-directory build-dir t)
+          (with-temp-file stale (insert "stale"))
+          (should (equal (file-name-as-directory build-dir)
+                         (ghostel--make-module-build-dir dir)))
+          (should (file-directory-p build-dir))
+          (should-not (file-exists-p stale)))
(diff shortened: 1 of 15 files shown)


test/ghostel-module-test.el
```diff
@@ -351,99 +372,66 @@ module beside a stale sidecar (issue #256 follow-up B1)."
...
+(ert-deftest ghostel-test-module-compile-command-uses-zig-build-prefix ()
   "Interactive compilation uses zig build directly."
   (let ((compile-invocation nil)
+        (finish-args nil)
         (default-directory nil)
         (ghostel-module-directory nil))
     (let ((native-comp-enable-subr-trampolines nil))
       (cl-letf (((symbol-function 'ghostel--resource-root)
-                 (lambda () "C:/ghostel/"))
+                 (lambda () "/src/ghostel/"))
```
Vonfry, line 398: The old codes looks for windows. Is the new one can work on windows as well?

Vonfry (approved):
With a glance, LGTM. And it is tested on linux.

test/ghostel-module-test.el
emil-e: I'm not really sure why it did because we currently don't support Windows. But Windows support has a PR almost ready to merge, so we will work on polishing up the Windows support at that time as well.

dakra: jfyi, It was "C:" because that test was actually coming from the original windows PR (that we merge soon). We just split it and some release/download stuff+tests got merged first.

PR #520, opened 2026-07-08
Reject input-mode commands outside ghostel buffers

### LINKED ISSUE #518: ghostel-readonly-fast-exit :set leak ghostel's readonly keymap into non-ghostel buffers
**version** 
ghostel 20260707.1804

**Summary** 
ghostel can install its readonly/copy keymap (and set ghostel--input-mode to copy/emacs) on buffers that are not ghostel-mode buffers. 

**issue**

- ghostel-readonly-fast-exit :set (ghostel.el:714, loop at :731-735) walks the entire (buffer-list) and re-installs the ghostel readonly
  keymap on every buffer whose ghostel--input-mode is copy/emacs:

dakra:
Fixes #518.

`M-x ghostel-copy-mode` (or any of the other input-mode switch commands) in a non-ghostel buffer ran the full mode-entry path: it replaced the buffer's local keymap with a ghostel keymap via `use-local-map`, set buffer-local `ghostel--input-mode` to `copy`/`emacs`, clobbered `cursor-type`, and installed the fake-cursor pre-redisplay hook. The major-mode keymap was gone until the mode was reinitialized.

Root cause: `ghostel--input-mode` is a `defvar-local` whose global default is `semi-char`, so every buffer looks like a ghostel buffer at rest, and none of the commands checked the major mode.

### Changes

- New `ghostel--ensure-ghostel-buffer` helper (same predicate and `user-error` message as the existing inline guards in `ghostel-send-string`/`ghostel-send-key`/`ghostel-paste-string`, which now use it too).
- Guard `ghostel-semi-char-mode`, `ghostel-char-mode`, `ghostel-copy-mode`, `ghostel-emacs-mode`, and `ghostel-line-mode` with it.
- Two regression tests in `test/ghostel-modes-test.el`: the five commands signal `user-error` in a non-ghostel buffer and leave the local keymap and `ghostel--input-mode` untouched, and the `ghostel-readonly-fast-exit` custom setter finds nothing to touch outside ghostel buffers (the issue's second path — with entry guarded, the copy/emacs state its buffer-list walk matches is unreachable there, so the setter needs no guard of its own).
- Two release-handler unit tests in `test/ghostel-mouse-paste-test.el` drove the real `ghostel-copy-mode` in plain temp buffers (the leak happening inside the test suite); they now mock it like their sibling tests.

### Testing

- `make -j8 all` green (build, lint, elisp + native + zig + evil test groups).
- Live-verified in a real session: all five commands error cleanly in `*scratch*` with the keymap intact, and copy/emacs/char/semi-char switching plus `q` fast-exit behave unchanged in a live zsh terminal.


emil-e (approved):

dakra:
merged into main

PR #522, opened 2026-07-08
Resolve terminal from WINDOW's buffer in `ghostel--adjust-size'

jcmfernandes:
## Problem

Buffer-local `window-size-change-functions` are run by redisplay with an
arbitrary buffer current, not necessarily WINDOW's buffer.
`ghostel--adjust-size` read `ghostel--term`/`ghostel--process` from the
ambient current buffer, so when it ran with some other buffer current —
e.g. a split from another window, or any programmatic resize — those
locals were nil, the `when` guard bailed, and the resize was silently
skipped. The grid stayed at its old size.

## Fix

Wrap the body in `with-current-buffer` on `(window-buffer window)` so the
terminal state is always resolved from WINDOW's own buffer, regardless of
what buffer happens to be current when the hook fires.

## Testing

`make test` — `ghostel-terminal-test` passes 16/16, including a new
regression test `ghostel-test-resize-window-buffer-not-current` that
drives `ghostel--adjust-size` with an unrelated buffer current and asserts
the resize still happens.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -4852,32 +4852,36 @@ If WINDOW was anchored to the live viewport before the size change,
 keep it anchored.  Redraw synchronously when the terminal size
 actually changes.  When FORCE is non-nil, run the resize/redraw
 path even when the row/column count is unchanged."
-  (when (and ghostel--term ghostel--process (ghostel--terminal-live-p))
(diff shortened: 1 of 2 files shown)


emil-e (approved):

emil-e:
Nice catch, thanks for the PR.

PR #528, opened 2026-07-11
Redraw rows from dirty terminal pages

kiennq:
Honor libghostty page-level dirtiness after full-viewport row rotation, and clear it only after rendering the page. Add coverage for alternate and no-scrollback primary screens.

Currently the Copilot CLI is affected. When row rotation happens, only the last row is updated.


emil-e:
Nice catch! Used a slightly different approach.

PR #532, opened 2026-07-13
ghostel-kitty: ensure scale of kitty images to avoid cropping

pank:
Emacs tries to match the height of kitty images to the surrounding text is scale isn't set due to `create-image`:

> create-image is an autoloaded native-comp-function in ‘image.el’.
> ...
> If the property ‘:scale’ is not given and the display has a high
> resolution (more exactly, when the average width of a character
> in the default font is more than 10 pixels), the image is
> automatically scaled up in proportion to the default font.

without the patch:
image

with the patch:
image

this is using ghostel v0.43.0 on windows.

### DIFF
--- a/lisp/ghostel-kitty.el
+++ b/lisp/ghostel-kitty.el
@@ -189,6 +189,7 @@ user shouldn't have to trigger one)."
                  (img (create-image data (if is-png 'png 'pbm) t
                                     :width (* g-cols cw)
                                     :height (* g-rows ch)
+                                    :scale 1
                                     :ascent 'center))
                  (skip (max 0 abs-row))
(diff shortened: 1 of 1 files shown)


emil-e:
Thanks for the PR. The kitty graphics support needs a lot of love, but this at least fixes the immediate issue. My plan is to rewrite the code soon.

pank:
I will continue to keep an eye out and test various bits and pieces around this. it's very exciting!

But I mostly use these things on my windows machine, and it remains to be seen if I can make zig actually compile. ($corp laptop acts like it has seen a ghost whenever something ends in `.exe`)

PR #537, opened 2026-07-13
Kill the partially created buffer when terminal creation quits

dakra:
`ghostel--create` cleans up after `error` but `quit` is not a subtype of `error`, so a keyboard quit during initialization — e.g. answering `C-g` to a dir-locals prompt while `ghostel-mode` runs `hack-dir-local-variables` — skipped the cleanup and left behind an empty `ghostel-mode` buffer with no terminal and no process (a blank `*ghostel*<N>` husk that plain `M-x ghostel` never reuses, since it has no buffer identity).

The handler now catches `(error quit)`, killing the partial buffer before re-signaling.

The regression test signals `quit` from inside initialization and asserts no buffer survives; it uses `condition-case` directly because ERT's `should-error` only traps `error`. Verified to fail without the fix.


dakra:
merged into main

PR #543, opened 2026-07-14
Make evil-ghostel ESC routing discoverable

### LINKED ISSUE #539: Need more visibility on evil-ghostel-toggle-send-esc
scenario: I'm a new user, I've got ghostel setup in doom emacs (+evil mode). I fire up a ghostel terminal and start claude TUI, and then I'm trapped with no way to exit the session. I haven't felt this way since I first tried to quit Vim. 

a few ideas: 
1. Put this somewhere more prominently in the ghostel README, particularly in or near the `Input modes` and `Evil` sections
2. Add a README to evil-ghostel
3. if you want to get fancy with it, detect rage typing / frustration signals and give the user some guidance.

I think there's also a story here for the user that doesn't even have evil-ghostel installed, but in my case Doom's config took care of that for me. 

Happy to put up a PR if requested

dakra:
Closes #539

New evil/Doom users starting a fullscreen TUI (e.g. Claude Code with `/tui fullscreen`) felt trapped: with the default `auto` routing, alt-screen ESC goes to the app, so normal state is unreachable via ESC, the leader key is dead in insert state, and nothing pointed at the escape hatch.

### Discoverability (`febcb6d`)

- **One-time hint**: the first time `auto` routing sends ESC to an alt-screen app, the echo area shows `ESC sent to the terminal app (alt-screen); C-c C-r routes it to evil instead` (once per session).
- **`C-c C-r`** is bound to `evil-ghostel-toggle-send-escape`, now a two-state toggle: from `auto` it switches to whichever mode differs from auto's current effect (`evil` while an alt-screen app runs, `terminal` otherwise); from an explicit mode back to `auto`. Every press visibly changes the routing — previously, cycling `auto` → `terminal` in an alt-screen app changed nothing observable, which is exactly what confused users in #539. Numeric prefixes 1/2/3 still set a mode directly, and landing on `auto` messages the effective routing (`auto (now → terminal)`).
- **`C-c <escape>`** runs `evil-force-normal-state` as a one-shot switch that leaves the routing untouched. It's bound on the `<escape>` function key, so it works on GUI frames and on ttys speaking the kitty keyboard protocol (via [kkp.el](https://github.com/benjaminor/kkp)); on legacy ttys the event never occurs, so it can't shadow the `C-c M-…` bindings.
- **Docs**: new *ESC routing* section in the manual, a note in the main README, a new `extensions/evil-ghostel/README.md`, and an expanded package commentary.

### Alt-screen detection cleanup (`d206461`)

Three different alt-screen checks existed: the native `ghostel--alt-screen-p` (reads the active screen), line-mode's 1049-or-1047 mode query, and evil-ghostel's 1049-only query. The mode queries miss alt screens entered via other DEC modes — e.g. `auto` ESC routing kept ESC in evil for a TUI using mode 1047.

A new public `ghostel-alt-screen-p` wraps the native check and is now used for line-mode pause/resume and all evil-ghostel alt-screen decisions; tests stub the native function instead of `ghostel--mode-enabled`. The extension `.elc` Make rule also gained a dependency on the core `.elc` files so parallel builds can't compile evil-ghostel against a stale `ghostel.elc`.

### Testing

- `make -j8 all` (build, elisp + native tests, lint) and `make test-evil` (130/130) pass.
- Live-verified in real Emacs sessions (tty and GUI): the trapped-user journey from #539, the hint, the toggle in and out of alt-screen, `C-c <escape>` on a GUI frame, and the mode-1047/47 coverage of the new predicate (where the old flag check provably returned nil).


SpyrosRoum:
Hey, anything stopping this from being merged?

dakra:
> Hey, anything stopping this from being merged?

Nothing. It was just not on my radar anymore.

Thanks for the reminder. I'll merge it later today.

PR #556, opened 2026-07-22
Fixes for bash w/ undefined variable usage & builtin type

zenspider:
- **Fixes for undefined variables in bash**
- **Fixed idempotency guard use of builtin type**


dakra:
Thanks. Merged.

PR #563, opened 2026-07-24
Upgrade to latest libghostty and Zig 0.16

emil-e:
## Summary

- upgrade the libghostty dependency and require Zig 0.16.0
- migrate the native module to the Zig 0.16 I/O, process, C translation, and synchronization APIs
- update CI, release builds, and source-build documentation for Zig 0.16.0

## Verification

- clean-checkout `zig build`
- clean-checkout `zig build test`
- native process tests: 5 passed
- native render tests: 83 passed
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Install Zig
         uses: mlugg/setup-zig@v2
         with:
-          version: 0.15.2
+          version: 0.16.0
 
       - name: Run pure-Zig unit tests
         shell: bash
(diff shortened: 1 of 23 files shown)


kiennq:
@emil-e Would it be okay to run benchmark between zig 0.15.2 and the new 0.16 change?
I heard Zig 0.16 slow down the perf on IO which can be the hot path for this ghostel

emil-e:
@kiennq Most of the I/O here doesn't go through the Zig standard library in the first case. In Zig 0.16, many of the POSIX helpers have been deleted even, so this branch is even more raw. On Windows, it's the same raw Windows APIs. On Windows in particular, there are throughput bottlenecks in ConPTY that have nothing to do with Zig so I doubt there's a difference there.

One thing that's a slight regression is that libghostty is now even faster since it uses SIMD for plaintext handling. This means that on POSIX, we receive more EAGAIN which makes it ever so slightly slower in raw throughput.

I asked my agent to run some comparison benchmarks and here's the AI slop it wrote out, if you're interested.
https://gist.github.com/emil-e/4372c5b0b78d723c3555481bf7021182

We're going to hold off a little bit on merging this until we have some runtime on it.

PR #571, opened 2026-07-28
Fix byte-compile issues by using load-prefer-newer

zenspider:
Found while trying to roll back some changes. Source files got updated
after byte compilation and recompiling outputs the following warnings:

```
emacs --batch  -Q -L lisp --eval "(setq byte-compile-error-on-warn t)" -f batch-byte-compile lisp/ghostel.el
Source file ‘.../lisp/ghostel-line-mode.el’ newer than byte-compiled file; using older file
```

Use load-prefer-newer to force byte-compile to use newest .el file,
rather than the older .elc file.


dakra:
Merged. Thanks

PR #587, opened 2026-07-31
Linkify wrapped paths anywhere in scrollback, not just the viewport

### LINKED ISSUE #582: continuing problems with hard newline
the following terminal output occurred:
```
3794 $ CI=1 ./bin/rails test
/Users/ryandavis/Code/monorepo/shared_gem/lib/shared_gem/semantic_logger/appender/base_semantic_appender.rb:4:in '<module:Appender>': uninitialized constant SemanticLogger (NameError)
```
but wraps as:
```
3794 $ CI=1 ./bin/rails test
/Users/ryandavis/Code/monorepo/shared_gem/lib/shared_gem/semantic_logger/appender/
base_semantic_appender.rb:4:in '<module:Appender>': uninitialized constant SemanticLogger (NameError)
```
which resulted in the path not being highlighted/clickable, and ffap only went to the directory.

dakra:
Fixes #582.

Paths that the terminal soft-wraps across two rows were not clickable and not
even mouse-highlighted in `rails test` output. Two independent causes, one
commit each.

## Scan coverage

`ghostel--schedule-link-detection` queued only the live viewport — `point-max`
minus `ghostel--term-rows` lines. Between two queue events a flood materializes
far more rows than that, and everything above the new viewport was never
scanned by any cycle, because nothing rescans. Those rows carry no `help-echo`,
`mouse-face` or `keymap`, which is exactly the reported "not clickable, doesn't
even highlight" — and it also defeats the `ghostel-previous-hyperlink`
workaround from the issue thread, since that searches for the `help-echo` a
scan was supposed to have applied.

The renderer already knows which buffer range it rewrote, so it publishes that
instead. Every character of terminal output is inserted by `flushSpan`, which
makes coverage gapless by construction. It also closes the cursor-line skip for
free — `isRowDirty` repaints the row the cursor left, so a match skipped as
active input is republished and rescanned — and it is cheaper than before when
output is slow: a two-row repaint queues two rows, not the whole viewport.

Supporting changes the new region size forces:

- The queued bounds become markers. `evictScrollback` deletes from the buffer
  start on almost every redraw during a sustained flood, which rots integer
  bounds queued 0.1 s earlier.
- The scan drains at most `ghostel--link-detection-chunk` characters per tick,
  taken from the newest end, so one redraw materializing megabytes no longer
  blocks Emacs for as long as scanning all of it takes, and rows on screen are
  linkified before the backlog. Re-arming always goes through a timer, so
  `ghostel-plain-link-detection-delay` of 0 cannot recurse through it.
- Process exit no longer cancels a partly drained scan: the program that
  flooded has exited, so nothing will ever repaint that text again.

## Row limit

`ghostel--wrap-joined-region` counted joined rows against
`ghostel--soft-wrap-row-limit` cumulatively across the whole scan region
instead of restarting at each hard newline, so a region holding more wrapped
lines than the limit stopped joining at every 50th one and lost that line's
link. Harmless while scans covered ~34 rows; load-bearing once a scan covers
whatever the renderer repainted.

## Testing

`make -j8 all` passes at both commits, so the first is not a broken
intermediate.

New elisp tests (no module): per-logical-line row limit, chunked drain reaching
the whole pending range, bounds surviving front deletion, zero delay re-arming
through a timer instead of recursing. New native tests: `ghostel--repainted-region`
publication, a regression test writing far more than `term-rows` lines in one
write, and pending-wrap completion across two redraws. Each new test was
confirmed to fail against the pre-fix code.

Verified live under elate as well — bash in a ghostel buffer, `cat` of a
2000-line log with 200 wrapped 121-character paths. Post-fix: 200/200 linkified
across three runs, every link carrying the full joined target, and
`ghostel-previous-hyperlink` reaching all 200 from `point-max`. Isolating the
two bugs against that harness gives 3 misses per run at identical offsets for
the row-limit bug, and 192–193 of 200 missing for the coverage gap at
`ghostel-plain-link-detection-delay` 0. Revealing a buffer that flooded while
hidden scans the visible rows on the first tick and walks the backlog down over
the next seven.

## Known follow-ups, not addressed here

- Line mode passes `full=t` on every redraw (to rebuild the prompt row after the
  input snapshot), which erases and re-renders the entire scrollback — 3.1 ms at
  a saturated default scrollback versus 0.005 ms incremental. That now queues the
  whole buffer for scanning too. It is correct rather than wasteful, since a full
  redraw wipes every detected link (measured: 40 → 0), which the old viewport-only
  scan never restored outside the visible rows — but the root fix is to stop
  forcing full redraws in line mode.
- `ghostel-compile--commit-pending-frame` publishes a region nothing consumes, so
  a compile buffer's final frame is never link-scanned. Pre-existing.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -964,6 +964,7 @@ Matches Ghostty 1.2.0's `bold-color' configuration."
                (let ((inhibit-read-only t)
                      (inhibit-modification-hooks t))
                  (ghostel--redraw ghostel--term t t))
+               (ghostel--schedule-link-detection)
                (ghostel--apply-cursor-style))))))
 
 (defvar-local ghostel--cursor-pos nil
@@ -1699,6 +1700,7 @@ restoring the terminal contents, with point realigned to the VT cursor."
     (if (> old-len 0)
         (progn
           (ghostel--redraw ghostel--term t t)
+          (ghostel--schedule-link-detection)
           ;; The reverted edit must not move point either; realign it.
           (when ghostel--cursor-char-pos
             (goto-char ghostel--cursor-char-pos)))
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -992,6 +992,12 @@ otherwise hide its cursor for whatever buffer it shows next.")
 Values match libghostty's cursor style enum: 0=bar, 1=block,
 2=underline, 3=hollow-block, or nil for hidden.")
 
+(defvar-local ghostel--repainted-region nil
+  "Buffer range the last redraw rewrote, as a (MIN . MAX) cons, or nil.
+Published by the renderer, which inserts every character of terminal
+output.  Every redraw that renders sets it: nil when it painted
+nothing, so the value never outlives the redraw that produced it.")
+
 (defvar-local ghostel--input-mode 'semi-char
   "Current input mode.
 One of `semi-char', `char', `copy', `emacs', or `line'.  See
@@ -3629,11 +3635,6 @@ EVENT is the state-change description passed by Emacs."
   (let ((buf (process-buffer process)))
     (when (buffer-live-p buf)
       (with-current-buffer buf
-        (when ghostel--plain-link-detection-timer
-          (cancel-timer ghostel--plain-link-detection-timer)
-          (setq ghostel--plain-link-detection-timer nil
-                ghostel--plain-link-detection-begin nil
-                ghostel--plain-link-detection-end nil))
         (ghostel--cancel-password-confirm-timer)
         (ghostel--spinner-stop)
         (remove-hook 'pre-redisplay-functions #'ghostel--fake-cursor-update t)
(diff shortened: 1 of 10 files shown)


lisp/ghostel.el
```diff
@@ -3629,11 +3634,10 @@ EVENT is the state-change description passed by Emacs."
...
       (with-current-buffer buf
-        (when ghostel--plain-link-detection-timer
-          (cancel-timer ghostel--plain-link-detection-timer)
-          (setq ghostel--plain-link-detection-timer nil
-                ghostel--plain-link-detection-begin nil
-                ghostel--plain-link-detection-end nil))
+        ;; The queued link scan drains over several ticks, and a program
+        ;; that floods and then exits leaves a backlog nothing will ever
+        ;; repaint.  Let the drain finish; it stops by itself once the
+        ;; buffer is gone.
```
emil-e: This is a comment on code that doesn't exist anymore, so I don't think the comment is needed at all.

src/Renderer.zig
```diff
@@ -191,6 +202,10 @@ pub fn redraw(self: *Self, env: emacs.Env, force_full: bool, force_sync: bool) !
         screen.pages.rows
     else
         screen.pages.total_rows;
+
+    if (self.repainted) |r| {
+        env.set("ghostel--repainted-region", env.cons(r.min, r.max));
```
emil-e, line 207: I think this should be set to `nil` if `self.repainted` is `null`. Otherwise, it's a bit inconsistent that it overwrites the previous value only sometimes.

emil-e (changes requested):

Issue #592, opened 2026-08-03
Lone Escape isn't delivered to the child under TTY Emacs (breaks Kitty-keyboard apps like Claude Code)


eddof13:
### Summary
Under **GUI Emacs**, pressing <kbd>Escape</kbd> in a ghostel buffer works: the physical key arrives as the `escape` event, which `ghostel--define-terminal-keys` binds to `ghostel--send-event`, and ghostel's mode-aware encoder sends the right bytes (`\e[27u` when the child has the Kitty keyboard protocol on, bare `\e` otherwise).

Under **TTY Emacs (`emacs -nw`)**, a lone <kbd>Escape</kbd> never reaches ghostel at all. Emacs claims the ESC byte (`0x1b`) as the **Meta prefix**, so it is held/decoded before any buffer keymap sees it, and the `escape` event is never generated. The keystroke is silently lost.

### Impact
Agents like **Claude Code** enable the Kitty keyboard protocol and then expect Escape as `\e[27u`. In TTY Emacs, "press Esc to close" in their modals (e.g. `/config`) does nothing, because the keypress never leaves Emacs. `ghostel-send-key "escape"` and `ghostel-send-next-key` work around it, but there is no bindable command and no way to use the actual key.

### Root cause / prior art
This is the classic Emacs terminal ESC-vs-Meta problem. `term.el`, `eat.el`, and `vterm` all bind ESC as a prefix sub-keymap (skipping `[`/`O` so CSI/SS3 function keys still decode) — **none** deliver a lone TTY Escape as a bare key; `eat` even requires pressing Escape twice. ghostel already does the hard part (the mode-aware encoder); the only gap is *TTY delivery* of the `escape` event.

### Proposed fixes (two PRs to follow)
1. **`ghostel-send-escape` command** — a small interactive command in the existing `ghostel-send-C-*` family, routing through the encoder. Gives a reliable, bindable Escape (explicit chord; does not reclaim the physical key). *Low risk.*
2. **Opt-in TTY escape decoder** (`ghostel-tty-escape-mode`, default off) — an `input-decode-map` `:filter` using a short `sit-for` (the same technique as evil-mode's `evil-esc-mode`): a lone ESC with no following byte becomes the `escape` event and flows through the existing encoder; a real Meta/CSI sequence is left untouched. Makes the **bare Esc key** work in TTY. Tradeoff: ~10 ms latency on the Meta prefix inside ghostel terminal-input modes, hence opt-in.

[...]

PR #593, opened 2026-08-03
Add ghostel-send-escape command

eddof13:
Part of #592.

Adds an interactive `ghostel-send-escape` alongside the existing `ghostel-send-C-*` commands. It routes Escape through the mode-aware key encoder, so the bytes are `\e[27u` when the foreground program has the Kitty keyboard protocol enabled and a bare `\e` otherwise.

**Why:** a bindable, reliable Escape. Handy under TTY Emacs, where a lone Esc is claimed as the Meta prefix and never reaches the terminal, so "Esc to close" in agents like Claude Code does nothing. Equivalent to `(ghostel-send-key "escape")`, but a proper interactive command in the send-key family. Left intentionally unbound — happy to add a binding wherever you'd prefer.

**Tests:** `ghostel-test-send-escape` in `test/ghostel-keys-test.el` (mirrors `ghostel-test-send-event`) asserts it hands `"escape"`/`""` to the encoder.

Sibling PR adds an opt-in `ghostel-tty-escape-mode`; both reference #592.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1806,6 +1806,19 @@ overlay clears the way \\`keyboard-quit' would in other buffers."
   (deactivate-mark)
   (ghostel--send-encoded "g" "ctrl"))
 
+(defun ghostel-send-escape ()
+  "Send Escape to the terminal, encoded for its current keyboard mode.
+Routed through the key encoder, so it is `\\e[27u' when the foreground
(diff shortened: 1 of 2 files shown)


dakra:
Thanks. But closing this in favor of https://github.com/dakra/ghostel/commit/6fe0b0624f02d199ef07b52cae62300fe3612701

PR #594, opened 2026-08-03
Add opt-in ghostel-tty-escape-mode for a lone Escape under TTY

eddof13:
Part of #592.

Adds a global, **off-by-default** `ghostel-tty-escape-mode` (and `ghostel-tty-escape-delay`) so the physical Escape key works in ghostel buffers under **TTY Emacs**. In a TTY, Emacs claims the ESC byte as the Meta prefix, so a lone Esc never reaches ghostel's `<escape>` binding and is lost.

**Mechanism:** the same `input-decode-map` `:filter` + `sit-for` trick as `evil-esc-mode`. A lone ESC not followed by more input within `ghostel-tty-escape-delay` is delivered as the `escape` event and flows through the existing encoder (`\e[27u` under Kitty, `\e` otherwise); a real Meta (ESC+key) or CSI/SS3 sequence has input already pending, so the filter passes the original map through and Emacs decodes it normally. Installed per text terminal (guarded by a terminal-parameter), cleaned up on disable, no-op under GUI.

**Tradeoff:** ~`ghostel-tty-escape-delay` (10 ms) latency on the Meta prefix inside ghostel terminal buffers — hence opt-in.

**Tested:**
- `test/ghostel-tty-escape-test.el` covers the filter branch logic (lone ESC → `escape`; Meta / CSI / non-ghostel → passthrough) and the mode's frame-hook lifecycle.
- Manually in Emacs 30.2 `-nw`: enabling it makes the bare Esc key close Claude Code's Kitty-protocol modals, with `M-x` and arrow keys intact.

Sibling to the `ghostel-send-escape` PR; both reference #592.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1806,6 +1806,84 @@ overlay clears the way \\`keyboard-quit' would in other buffers."
   (deactivate-mark)
   (ghostel--send-encoded "g" "ctrl"))
 
+
+;;; Lone Escape under TTY Emacs
+
+;; In a TTY, Emacs claims the ESC byte as the Meta prefix, so a lone Esc
+;; never reaches ghostel's `<escape>' binding and is lost.  `ghostel-tty-
+;; escape-mode' translates a bare ESC into the `escape' event so it flows
+;; through the normal encoder, using the same input-decode trick as
+;; `evil-esc-mode'.  Off by default because of the Meta-prefix latency.
+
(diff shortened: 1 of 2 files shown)


dakra:
Thanks. But closing this in favor of https://github.com/dakra/ghostel/commit/6fe0b0624f02d199ef07b52cae62300fe3612701

PR #607, opened 2026-08-06
Fix stale [Compiling] modeline and copy-mode trap in ghostel-compile

### LINKED ISSUE #605: Killing compilation buffer before finishes mantain the [compiling] status in modeline and popper interation
Hi, i trying this package and found two strange behaviour, with ghostel-compile-global-mode or using ghostel-compile, killing compilation buffer before it's finish mantain the [compiling] in the modeline.

The second behaviour is, using popper, if i kill the buffer using `popper-kill-latest-popup` to kill the compilation buffer, the next time there i run compile, the buffer opens showing nothing and this message appears in the minibuffer: `Copy mode: press any printable key to exit`, clicking any key, show the content of the compilation buffer, but the keybinds are like the ghostel-mode, ie, `C-p` is `ghost--send-event`, not `previous-line`, only returns to the correct keybinds when the compilation finish.

edit:
Emacs version: 32.0.50

dakra:
Fixes #605.

Two bugs in compilation-style `ghostel-compile` buffers:

1. **Stale `[Compiling]` modeline**: killing a compile buffer mid-run fired the sentinel with a dead buffer, which skipped the `compilation-in-progress` cleanup — the dead process stayed in the list and the global `[Compiling]` indicator stuck forever.
2. **Copy-mode trap under popper**: the deferred point-leave check after the `M-x ghostel-compile` minibuffer exit fired in the popper-selected popup window and auto-entered copy mode (frozen terminal, no output). Exiting copy mode then restored `ghostel-semi-char-mode-map`, clobbering `ghostel-compile-view-mode-map` so `C-p`/`C-n`/`RET` were sent to the PTY until finalize.

One commit per fix:

- **Drain `compilation-in-progress` when the compile buffer dies mid-run**: the sentinel now runs the cleanup in an `unwind-protect` cleanup form, mirroring stock `compilation-sentinel` (including its ordering: finalize before the `delq`, so finish hooks observe stock-compatible state).
- **Suppress automatic copy-mode entry in compilation-style buffers**: the point-leave check and mark activation are gated on `ghostel--inhibit-insert-forwarding`, the existing compilation-style marker. Interactive compile runs (flag nil) keep auto-switching.
- **Restore compilation-style state when exiting copy or Emacs mode**: new buffer-local `ghostel--readonly-exit-function`; when set, `ghostel-readonly-exit` delegates the post-exit restore to it and the `*-and-send` exits don't forward the triggering key. Compile buffers install `ghostel-compile--restore-compilation-style` (view keymap, read-only barrier, `:run` indicator, redraw-timer restart), set/cleared in lockstep with the inhibit flag through a shared setter. This also covers the remaining paths into copy mode there (explicit `M-x ghostel-copy-mode`, link follow).

Tests: 8 new (6 elisp, 2 native), including an end-to-end kill-mid-run test and an integration test that goes through `ghostel-compile--start` → `ghostel-copy-mode` → exit against the production wiring.

Verified live (Emacs 31, real popper 0.4.8): kill mid-run leaves no `[Compiling]`; with popper the compile popup opens in semi-char with output streaming and view keys bound; copy-mode exit restores the view state with no byte sent to the PTY.

### DIFF
--- a/lisp/ghostel-compile.el
+++ b/lisp/ghostel-compile.el
@@ -562,29 +562,34 @@ destroys the grid, so commit it here or lose the output it holds."
 (defun ghostel-compile--sentinel (process _event)
   "Sentinel for the compile PROCESS: finalize the buffer on exit."
   (when (memq (process-status process) '(exit signal))
-    (let ((buffer (process-buffer process))
-          (exit (process-exit-status process)))
-      (when (buffer-live-p buffer)
-        (with-current-buffer buffer
-          (when ghostel-compile-debug
-            (message "ghostel-compile: sentinel exit=%S status=%S"
-                     exit (process-status process)))
-          ;; Cancel any scheduled redraw and commit the current terminal state
-          ;; to the buffer synchronously.  Without this, a short-lived
-          ;; command (`echo`, `false`, `exit 7`) finishes before the
-          ;; ~16 ms redraw timer fires and its output is lost when
-          ;; `--teardown-terminal' destroys the renderer.
-          (when ghostel--term
-            (when ghostel--redraw-timer
-              (cancel-timer ghostel--redraw-timer)
-              (setq ghostel--redraw-timer nil))
-            (ghostel--redraw-now buffer)
-            (ghostel-compile--commit-pending-frame buffer))
-          (setq compilation-in-progress
-                (delq process compilation-in-progress))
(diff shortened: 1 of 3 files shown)


BrunodaSilvaBelo:
I forgot to mention, my emacs version is 32.0.50

PR #617, opened 2026-08-11
Don't conflate equally named projects in project buffer matching

### LINKED ISSUE #432: ghostel-project incorrectly reuses terminal buffers for same-named projects in different directories.
Steps to Reproduce

  Create two separate project directories with the same name but different paths, for example: ~/a/b and ~/d/b. Ensure project.el recognizes them as distinct projects (e.g., by placing a .git or .project file in each).
 
   1  Open a file in the first project: ~/a/b/c.md.
   2  Run M-x ghostel-project. This creates a terminal buffer (e.g., *ghostel:b*) with default-directory set to ~/a/b/.
   3  Open a file in the second project: ~/d/b/c.md.
   4  Run M-x ghostel-project again.

Expected Behavior

A new terminal buffer should be created (or an existing one specific to ~/d/b should be opened) with default-directory correctly set to ~/d/b/.

Actual Behavior

Emacs switches to the existing terminal buffer created in step 3, which is stuck in ~/a/b/. The terminal for ~/d/b is never created. Furthermore, commands like ghostel-project-next and ghostel-project-previous will also conflate the two projects.

mrcnski:
Fixes #432.  (The issue was closed without a linked fix. The local same-name case it describes still reproduces on `main`.)

## Problem

`ghostel-project` and the project-scoped buffer commands identify a project's buffers by `ghostel--buffer-identity`, which is the project-prefixed buffer name, e.g. `*<root-dirname>-ghostel*`.

The name alone is ambiguous: two projects that share a directory name (say `~/work/website` and `~/oss/website`) produce the identical identity `*website-ghostel*`, so `ghostel-project` in project A switches to project B's terminal instead of creating one for A. The `identity` (and default `both`) scope of `ghostel--project-buffers` lists and cycles B's buffers while you are in A.

#499 fixed the sibling collision, a local and a remote root with the same path, by folding the TRAMP host into the name, but two local (or two same-host) projects with the same directory name still collide.

## Fix

Record the project root at creation and key identity matching on it.  **This mirrors how magit keys its per-repository buffers**: reuse compares `(major-mode, magit--default-directory, lock value)`, a toplevel recorded at creation, and never the buffer name, which stays purely cosmetic.

- New buffer-local `ghostel--buffer-project-root`, set at creation from the new dynamic variable `ghostel--project-root`, which `ghostel-project` binds around its `ghostel` call.  Plain `ghostel` buffers record nil.
- Identity-scope matching in `ghostel--project-buffers` keys on the recorded root, regardless of name.  **This also closes a pre-existing gap**: numbered instances (`*website-ghostel*<2>` from a numeric-arg `ghostel-project`) never matched the `equal` comparison against the un-numbered prefix, so identity scope missed them once their shell cd'd out of the project.
- Buffer reuse (`ghostel--find-buffer-by-identity`) additionally requires the roots to agree when both sides are non-nil. The name still distinguishes numbered instances within one project, so `C-5 M-x ghostel-project` finds instance 5 specifically.
- A nil recorded root (buffers from older sessions, `ghostel-bookmark` restores, or manual `ghostel-buffer-name` binders) keeps matching by name alone, so existing sessions and workflows are unaffected.

The displayed buffer name is intentionally unchanged. This fixes the matching key only.  (Uniquifying the visible name on conflict, e.g.  `*website-ghostel*` vs `*website<oss>-ghostel*`, would be a separate, purely cosmetic follow-up.)

## Other notes

Four new tests across `test/ghostel-buffers-test.el` and `test/ghostel-project-test.el`.

CHANGELOG.md updated under [Unreleased].

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,14 @@ All notable changes to this project will be documented in this file.
   t.  Fixes [#599](https://github.com/dakra/ghostel/issues/599).
 
 ### Fixed
+- `ghostel-project` and the project-scoped cycling commands no longer
+  conflate projects that share a directory name.  The identity used
+  for buffer reuse and for `identity`-scope matching was the
+  project-prefixed buffer name alone, so two projects both named e.g.
+  `website` shared buffers.  Buffers now record their project root at
+  creation (mirroring how magit keys buffers on the repository
+  toplevel rather than the buffer name).  Buffers from older sessions
+  (no recorded root) keep matching by name alone.
 - evil-ghostel: typing the first key of an `evil-escape` chord (e.g.
   `jk`) no longer doubles the character in the shell, and completing
   the chord no longer leaks it.  evil-escape previews the first key
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -389,10 +389,12 @@ project:
   against the current project root.  Follows shell `cd'.
   A terminal that walked out of the project is excluded; a plain
   `ghostel' buffer that cd'd into the project is included.
-- `identity': match each buffer's `ghostel--buffer-identity'
-  against the name `ghostel-project' would use for the current project.
-  Stable across `cd', but only finds buffers originally
-  created via `ghostel-project'.
+- `identity': match buffers created via `ghostel-project' for this
+  project by their recorded project root, so equally named projects
+  stay separate and numbered instances are found.  Buffers without a
+  recorded root (older sessions) fall back to comparing
+  `ghostel--buffer-identity' against the project-prefixed name.
+  Stable across `cd'.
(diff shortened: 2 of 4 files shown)


mrcnski:
Note to self: I'm interested in contributing desktop integration which would rely on the `ghostel--buffer-project-root` in this PR. I'll wait for feedback on this PR, first. Leaving this note so I don't forget about that followup.

dakra:
Thanks for the PR.
I make PR #625 
Since we had a few troubles with buffer identity before I refactored it completely and made it a lot more robust and flexible.
`ghostel-identity` is now more granular and can hold more things then before where the buffer-identity was simply a string.
This should make your desktop.el integration also easier.

Issue #133, opened 2026-04-18
Does ghostel have something like, or work with multi-vterm?


ackerleytng:
Thanks for ghostel!

Does ghostel have something like, or work with multi-vterm?

dakra:
I haven't heard of multi-vterm yet. But had a quick look, and looks like just a simple buffer management around vterm.
Something like this should also be easily implementable for ghostel.

I saw on reddit recently https://github.com/CsBigDataHub/popterm.el and they support ghostel.
Maybe that's something for you?

ackerleytng:
Thanks for the tip, I'll look into that :)

Was indeed looking for some way to have multiple terminals in different windows.

dakra:
No problem.

I'll close this issue then because I don't think there's anything special that ghostel needs to support.

But feel free to reopen or report any other issues that you find.
ghostel is still very young.

PR #620, opened 2026-08-12
Add ghostel-project-dwim

mrcnski:
## Problem

Working with more than one terminal per project is awkward, because no single command is safe to bind as "take me to a project terminal":

- `ghostel-project` keys its reuse check on the unnumbered project identity. Once only numbered or renamed instances remain (e.g. open two project terminals, kill the first), it creates yet another terminal instead of switching to the surviving one.
- Reaching a specific one of several project terminals means `ghostel-project-next` cycling or `ghostel-project-list-buffers`, and both signal an error when the project has no terminal yet.

(Related: #133 asked for support for multiple terminals per project. The cycle and list commands have since been added to ghostel, but without a tidy way to make everything work together in a single command.)

## Fix

A new command, `ghostel-project-dwim`, that does the expected thing at every amount of project terminals:

- with no project terminal it creates one via `ghostel-project`
- with exactly one it switches to it, whatever it is named
- with several  it prompts via `read-buffer`, defaulting to the forward-cycle buffer like `ghostel-project-list-buffers` does. Membership uses the existing `ghostel-project-buffer-scope` machinery.
- A prefix argument passes through to `ghostel-project`, so `C-u` still forces a new terminal and numeric session arguments keep their meaning.

For what it's worth, I surveyed how comparable packages (project.el's `project-shell`/`project-eshell`, eat, vterm, multi-vterm, mistty, vterm-toggle, shell-pop) handle several-terminals-per-project. All of them provide numeric prefix indices or blind cycling, consistent with ghostel, but none supports a picker. As far as I could tell, **this would be a novel feature unique to ghostel**. 🚀

Since ghostel already ships a `read-buffer` picker, extending it into a dwim entry point seemed like the natural ghostel-native answer.

## Other info

Four new tests in test/ghostel-buffers-test.el cover the three branches and the prefix pass-through.

README command table and CHANGELOG entries included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file.
   `desktop-read` starts a fresh shell in that directory under the
   saved buffer name, so restored window configurations find the
   buffer.
+- New command `ghostel-project-dwim`.  It creates a terminal when the
+  project has none, switches to it when there is exactly one, and
+  prompts via `read-buffer` when there are several.
 
 ### Changed
 - The bookmark functions are now public: `ghostel-bookmark-make-record`
--- a/README.org
+++ b/README.org
@@ -1824,6 +1824,7 @@ Enable it for any Emacs Lisp input method:
 |--------------------------------------+---------------------------------------------------------------------------|
 | =M-x ghostel=                          | Open a new terminal (create a new buffer with a prefix arg).              |
 | =M-x ghostel-project=                  | Open a terminal in the current project root (new buffer with prefix arg). |
+| =M-x ghostel-project-dwim=             | Project terminal: create if none, switch if one, pick if several.         |
 | =M-x ghostel-other=                    | Switch to the next terminal or create one.                                |
 | =M-x ghostel-next=                     | Cycle to the next ghostel buffer (sorted by name, wraps).                 |
 | =M-x ghostel-previous=                 | Cycle to the previous ghostel buffer.                                     |
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -38,6 +38,7 @@
 ;;
 ;;   M-x ghostel               Open a new terminal
 ;;   M-x ghostel-project       Open a terminal in the current project root
+;;   M-x ghostel-project-dwim  Project terminal: create, switch, or pick
 ;;   M-x ghostel-other         Switch to next terminal or create one
 ;;   M-x ghostel-next / ghostel-previous
 ;;                             Cycle through all ghostel buffers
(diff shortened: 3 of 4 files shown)


dakra:
Thanks for the PR.

I'm in the process of overhauling the whole public API anyway.
I'll have to sleep a night or two over your PR.
At first glance it's not necessarily what I would assume a -dwim command does,
it's basically just `ghostel-project` but when there are more than 1 buffer in the project it's `ghostel-project-list-buffers`.

mrcnski:
Yep, it's basically a convenience wrapper over those two commands, picking the right one situationally. No worries if this isn't accepted, I'm just tinkering and contributing in case it's useful to others. :) Out of curiosity, what was your expectation for a dwim command?

In case it helps for the public API, here's my research into what's provided by other packages. 

| Package      | Named always-create?    | Force-new mechanism           | Several-terminals answer      |
|--------------|-------------------------|-------------------------------|-------------------------------|
| project.el   | no                      | `C-u`                         | — (reuses first match)        |
| eat          | no                      | `C-u` / numeric index         | numeric index                 |
| vterm        | no                      | `C-u` / numeric / string name | numeric index                 |
| multi-vterm  | base command always new | n/a                           | cycle (`-next`/`-prev`)       |
| mistty       | yes (`mistty-create`)   | prefix, but overloaded        | cycle (base command)          |
| vterm-toggle | no                      | delegates to vterm            | cycle (`-forward`/`-backward`)|
| shell-pop    | no                      | numeric index                 | numeric index                 |


- "Named always-create" means whether there's a dedicated function to always create a new project terminal. For ghostel the answer would be no, which follows the dominant convention (call the project command with C-u to force a new project terminal).
- For the final column ("Several-terminals"), I believe that ghostel combines prior art by accepting both numeric indices and cycling, which I think is sensible.

mrcnski:
I've updated my personal dwim command to build on the recently-added consult-ghostel functionality:


```el
  (defun ghostel-project-dwim (&optional arg)
    "Create, switch to, or consult-pick this project's ghostel terminal.
With prefix ARG, force a new terminal via `ghostel-project'."
    (interactive "P")
    (let ((bufs (and (not arg) (project-current nil)
                     (ghostel-project-buffer-list))))
      (cond ((or arg (null bufs)) (ghostel-project arg))
            ((null (cdr bufs))
             (pop-to-buffer (car bufs)
                            (append display-buffer--same-window-action
                                    '((category . comint)))))
            (t (consult-ghostel-project)))))
```

The result? Why it looks splendid (and works intuitively!).

CleanShot 2026-08-30 at 14 41 01@2x

dakra:
@mrcnski I'm thinking of changing the default behaviour of consult-ghostel(-project) to yours.
Or at least when there is no Terminal to just create one.
I don't think for consult-xxx commands there is a precedence like for the other terminal commands so it's not confusing other users (e.g. `M-x ghostel` has to behave like it does because `M-x term`, `M-x vterm` etc all behave the same).
What do you think?

mrcnski:
Thanks for considering this and sharing your thoughts.

Regarding `ghostel` keeping precedent with `term`, `vterm` etc.: completely agreed here. In case it wasn't clear, my proposal was not about affecting the behavior of existing commands but about adding a *new* command, named something like `ghostel-dwim`.

About the consult commands getting some (or all) of the dwim behavior: here I think we should look at what existing consult commands do, and opt for consistency. I had Fable look at them and it found that "no built-in [consult command] bypasses the minibuffer when there is exactly one candidate" and that "there is [...] a general consult precedent, and the current consult-ghostel already fits it".

So here I think the more ergonomic functionality should live in a wrapper around the default consult ghostel commands - maybe documented somewhere like the README if it's too opinionated to bundle with the package.

PR #623, opened 2026-08-12
Clear the terminal title on an empty OSC 0/2

### LINKED ISSUE #619: Empty OSC 0/2 title (title clear) never reaches elisp
## Problem

A terminal application or shell that *clears* its title (`printf '\e]2;\a'`, the standard way to reset a title set earlier) has no effect in ghostel. The clear never reaches elisp. `ghostel--title` keeps the stale value and `ghostel-buffer-name-function` is never called, so nothing driven by the title can react to it ending.

The elisp side already anticipates this event. The `ghostel-buffer-name-function` docstring documents that TITLE "may be nil or empty", and `ghostel-buffer-name-by-title` explicitly handles that case. The module just never delivers it.

## Repro

With a logging name function to make delivery observable:

```elisp
(setopt ghostel-buffer-name-function
        (lambda (title) (message "ghostel title: %S" title) nil))
```

1. `M-x ghostel`
2. `printf '\e]2;hello\a'`: logs `ghostel title: "hello"` (and with the default function, the buffer renames).
3. `printf '\e]2;\a'`: nothing is logged, and `ghostel--title` still holds `"hello"`.

Expected at step 3: the name function is called with nil or `""`, per its documented contract.

## Where it's dropped

`titleChangedCallback` in `src/handler.zig` only forwards non-null titles:

```zig
/// Called when the terminal title changes.
fn titleChangedCallback(handler: *gt.TerminalStream.Handler) void {
    const self: *Self = @fieldParentPtr("inner", handler);
    const title = handler.terminal.getTitle();
    if (title) |t| {
        self.context.effect("ghostel--set-title", .{t});
    }
}
```

That guard is the only drop point. libghostty delivers the clear correctly end to end (traced in the vendored 1.3.2-dev sources):

- The OSC parser starts its payload capture at the `;` (`osc.zig`), so an empty payload still emits `change_window_title = ""` rather than no command.
- The stream handler's `windowTitle` stores the title and fires the `title_changed` effect unconditionally (`stream_terminal.zig`).
- `Terminal.setTitle("")` clears the stored title, so the callback's `getTitle()` returns null (`Terminal.zig`), which the guard above then swallows.

## Fix (implemented and tested)

Forward the clear as the empty string, which the documented elisp contract already allows:

```zig
/// Called when the terminal title changes.  A cleared title (empty
/// OSC 0/2 payload) reads back as null from `getTitle` and is
/// forwarded as the empty string so the clear reaches elisp.
fn titleChangedCallback(handler: *gt.TerminalStream.Handler) void {
    const self: *Self = @fieldParentPtr("inner", handler);
    const title = handler.terminal.getTitle();
    self.context.effect("ghostel--set-title", .{title orelse ""});
}
```

Verified with the module test rig on both pty backends (emacs-pty and native-pty). Feeding set/clear via OSC 2+ST, then set/clear via OSC 0+BEL, delivers only `("hello" "world")` to elisp before the change and `("hello" "" "world" "")` after it, with `ghostel--title` ending empty instead of stale. `zig build test` and the native elisp suite pass with the change.

## Why it matters

- Shells commonly set the title in preexec and clear it in precmd. Because the clear is dropped, any title-driven buffer naming keeps claiming a command is running at an idle prompt.
- Full-screen apps that set a title and clear it on exit (Claude Code is where I was hitting this) leave their name on the buffer forever.
- It blocks the shell-side half of the naming scheme discussed in #568: "revert the name at the prompt" cannot be driven by the shell emitting a title reset, which is the natural place for it. My workaround is clearing Emacs-side on the OSC 133 D marker (posted in that thread), which works but shouldn't be necessary.

Happy to PR the handler change plus a regression test. One open question for that PR: should `ghostel--set-title` also normalize `""` to nil, so `ghostel--title` stays nil-or-nonempty for elisp consumers? The name-function contract already covers both, so it's cosmetic.

dakra:
An empty OSC 0/2 title (`printf '\e]2;\a'`, the standard way to reset a title) was silently dropped at the native title callback: libghostty stores a cleared title as *no title*, `getTitle` returns null for it, and the callback only forwarded non-null titles. The stale title stayed in the mode line and in a title-derived buffer name until the next title report.

## Changes

- **`src/handler.zig`** — `titleChangedCallback` forwards a cleared title as `""` instead of dropping it. libghostty delivers the clear correctly end to end (empty payload still parses as `change_window_title`, `title_changed` fires unconditionally); this guard was the only drop point.
- **`lisp/ghostel.el`** — `ghostel--set-title` treats an empty title as "as if no title was ever set", matching Ghostty and iTerm2: `ghostel--title` normalizes to nil (invariant is now nil-or-non-empty, so the mode line falls back to the plain buffer name), and a title-derived buffer name reverts to `ghostel--buffer-identity`. The revert only applies to names title tracking has claimed and still goes through the manual-rename guard.
- `ghostel--init-buffer` backfills a nil `ghostel--buffer-identity` with the creation name so buffers created outside the `ghostel` command (`ghostel-exec`, eshell visual commands) have a revert target.
- The now-dead empty-string handling in `ghostel--buffer-identification` and `ghostel-buffer-name-by-title` is removed; the `ghostel-buffer-name-function` contract is now "nil or non-empty string" (custom name functions see nil on a clear, never `""`).

## Tests

- New native PTY-matrix regression test: title set, then cleared via OSC 2+ST, OSC 2+BEL, and OSC 0+BEL — asserts nil `ghostel--title` and the reverted buffer name on both backends.
- New elisp unit tests: clear reverts a by-title name to the identity, declines after a manual rename, is a no-op for buffers title tracking never renamed, and leaves by-directory names alone.
- Verified live in a terminal Emacs session (bash `--norc`, shell integration off): OSC 0/2 set/clear round-trips update `ghostel--title`, the mode line (no empty-parens leftovers), and by-title renames/reverts, with manual renames respected.

Fixes #619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,14 +31,21 @@ All notable changes to this project will be documented in this file.
   t.  Fixes [#599](https://github.com/dakra/ghostel/issues/599).
 
 ### Fixed
+- An empty OSC 0/2 title (e.g. `printf '\e]2;\a'`) now clears the
+  terminal title instead of being ignored, matching how Ghostty and
+  iTerm2 treat it (reset to as if no title was ever set): the mode
+  line falls back to the plain buffer name, and a buffer renamed by
+  title tracking reverts to its original name (manual renames are
+  still respected).
+  Fixes [#619](https://github.com/dakra/ghostel/issues/619).
 - evil-ghostel: typing the first key of an `evil-escape` chord (e.g.
   `jk`) no longer doubles the character in the shell, and completing
   the chord no longer leaks it.  evil-escape previews the first key
   with a speculative insert it reverts moments later; since terminal
   buffers became writable (0.47.0) that preview was forwarded to the
   PTY where it could not be reverted.  The preview is now skipped in
-  ghostel buffers — the chord itself keeps working.  Fixes
-  [#597](https://github.com/dakra/ghostel/issues/597).
+  ghostel buffers — the chord itself keeps working.
+  Fixes [#597](https://github.com/dakra/ghostel/issues/597).
 
 ## [0.49.0] — 2026-08-02
 
@@ -55,8 +62,8 @@ All notable changes to this project will be documented in this file.
   command with no trailing newline gets its link once the prompt
   returns.  Soft-wrap joining also restarts its row limit at every
   hard newline; regions holding more than 50 wrapped lines used to
-  stop joining at each 50th, losing that line's link.  Fixes
-  [#582](https://github.com/dakra/ghostel/issues/582).
+  stop joining at each 50th, losing that line's link.
+  Fixes [#582](https://github.com/dakra/ghostel/issues/582).
 - Redraws that rebuild the whole buffer — changing `ghostel-bold-color`,
   the foreign-insert repair, line-mode teardown — no longer strip
   detected file/URL links for good; they queue a rescan of what they
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -407,7 +407,9 @@ project:
 (defcustom ghostel-buffer-name-function nil
   "Function returning the ghostel buffer name, or nil to leave it unchanged.
 Called in the ghostel buffer with one argument, the terminal TITLE (the
-OSC 2 string; may be nil or empty), on both a title change and a `cd' \(OSC 7).
+OSC 2 string; nil when the terminal has no title), on both a title change
+and a `cd' \(OSC 7).  A title clear (empty OSC 0/2) reverts a
+title-derived name to the buffer's original name.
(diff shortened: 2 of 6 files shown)


mrcnski:
Thanks for picking this up so quickly! I built the module from this branch and ran it on macOS (aarch64, Emacs 30). The new regression test passes here, and a set/clear round-trip now clears `ghostel--title` and reverts a by-title buffer name.

The handler.zig change matches what I found when filing #619. The identity backfill for `ghostel-exec` buffers giving every buffer a revert target is a good touch. My config's name function reads `ghostel--buffer-identity` and benefits from the same guarantee.

LGTM!

Issue #568, opened 2026-07-27
Don't change buffer name when running commands


krisbalintona:
Hello,

As I understand it, Ghostel buffers' names are changed to display some information about the current command running in that buffer. But this has presented an annoyance I don't observe in other packages, like EAT or Vterm, related to buffer name uniquification (see `uniquify-buffer-name-style` and friends).

To describe what I mean:
1. Open first Ghostel buffer, auto-named to `*ghostel: ~/*`.
2. Run `sleep 10` in that buffer.
3. Its buffer name is changed to `*ghostel: sleep 10 ~/*`.
4. Open a new Ghostel buffer with `C-u M-x ghostel`.
5. Its buffer name was the original (first) Ghostel buffer's name: `*ghostel: ~/*`.
6. First Ghostel buffer's name, after finishing `sleep 10`, is now named `*ghostel: ~*<2>` (according to my value of  `uniquify-buffer-name-style`).
You can see how this invites confusion about which Ghostel buffer I've been using for what.

I think a better solution would instead be to do what major modes like Info do and change not the buffer name but the buffer identification in the mode line. See `mode-line-buffer-identification`; for an example of how, see `Info-set-mode-line`.

dakra:
You can customize that with `ghostel-buffer-name-function`.

I know a few users who don't like this behavior, so maybe we should change the default?
Not sure yet as I personally like it.
Good tip with `mode-line-buffer-identification`

krisbalintona:
> You can customize that with `ghostel-buffer-name-function`.

Thanks for letting me know about this option.

> I know a few users who don't like this behavior, so maybe we should change the default? Not sure yet as I personally like it. Good tip with `mode-line-buffer-identification`

My take is that `ghostel-buffer-name-function` should exist but the default should be changed (to `nil`?) alongside setting `mode-line-buffer-idenfitication` buffer locally. For instance, all the Info manual buffers are named like `*info*<2>` but the mode line shows info on which manual one is in and what node.

zenspider:
FWIW, I have a function that renames my ghostel shell buffers to `shell-<N>` and they don't show anything while running a process.

dakra:
> FWIW, I have a function that renames my ghostel shell buffers to shell-<N> and they don't show anything while running a process.

I think that's a `fish` thing that they emit OSC 2 (title change) for switching directories and stuff.

krisbalintona:
[...]

PR #624, opened 2026-08-12
Annotate buffer picker candidates with the terminal title

mrcnski:
## Problem

Since the stable-names change (#568), buffer names no longer carry the terminal title. This is right for name stability, but it left the pickers blind.

With several terminals in one project, `ghostel-list-buffers` and `ghostel-project-list-buffers` complete over names like `*ghostel*` and `*ghostel*<2>` with nothing to say what each terminal is running. The mode line (`ghostel-buffer-identification-format`) only shows the title of the buffer you are already visiting, so it cannot help at pick time.

## Fix

`ghostel--read-buffer` now provides an annotation function through `completion-extra-properties`. Each candidate is annotated with its terminal title, capped at `ghostel-annotation-title-width` columns (a new defcustom, default 30 to match the `%.30t` in `ghostel-buffer-identification-format`). nil annotates with the full title. Terminals with no title (at a quiet prompt) simply get no annotation.

The annotation function, `ghostel-annotate-buffer`, is public so other completion frontends can reuse it. As a marginalia user, I found that marginalia replaces annotations for the whole `buffer` category with its own, which would hide these.

The README documents a small wrapper that prepends the title to `marginalia-annotate-buffer`. Documenting this seemed better than shipping code that rewrites marginalia's registry behind the user's back.

## Other info

Three new pure-elisp tests. README and CHANGELOG entries included.

On the default width: 30 matches the mode-line convention, but the picker has a whole line per candidate, so happy to bump the default if you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ All notable changes to this project will be documented in this file.
   `desktop-read` starts a fresh shell in that directory under the
   saved buffer name, so restored window configurations find the
   buffer.
+- The buffer pickers (`ghostel-list-buffers`, etc.) annotate each
+  candidate with its terminal title, so completion shows what every
+  terminal is running even though the stable buffer names no longer
+  carry it.  The title is capped at `ghostel-annotation-title-width`
+  columns (default 30, nil for the full title).  Marginalia users need
+  a small wrapper around `ghostel-annotate-buffer`, see the README.
 
 ### Changed
 - The bookmark functions are now public: `ghostel-bookmark-make-record`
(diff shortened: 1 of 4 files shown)


mrcnski:
How the annotations look for me with marginalia (default annotation width of 30):

CleanShot 2026-08-12 at 07 52 17@2x

This picker is from my `ghostel-project-dwim` command [added here](https://github.com/dakra/ghostel/pull/620). :)

dakra:
I'll have a look at this.
Slightly related is this over 2 week old consult branch if you haven't seen it: https://github.com/dakra/ghostel/commit/95073046387cd261cb0070416a5ac2fde1564ab7

I'll likely change it though, probably do a extra consult-ghostel package instead of bundling, that's why it's not merged yet.

mrcnski:
Nice! I'll rebase this PR so there are no conflicts when you review.

The consult branch looks good. I think it would be complementary to the changes here. My change covers the annotation for the built-in read-buffer pickers without conflicting with your branch. And a consult source could reuse my `ghostel-annotate-buffer` as its `:annotate`.

dakra:
Merged. Thanks ❤️

PR #627, opened 2026-08-12
Add desktop.el session persistence

mrcnski:
## Problem

Terminals don't survive an Emacs restart.  Even with `desktop-save-mode` enabled, every ghostel buffer is simply gone after `desktop-read`. desktop.el skips non-file buffers unless their mode opts in, so a saved session restores files and window layouts only.

#625 made a fix straightforward: the desktop handlers persist one `ghostel-identity` alist.

## Fix

A new `ghostel-desktop.el`, mirroring `ghostel-bookmark.el`'s architecture:

- `ghostel-mode` points the buffer-local `desktop-save-buffer` at an autoloaded save function (a quoted symbol, no load-time dependency)
- ghostel.el registers the autoloaded restore handler in `desktop-buffer-mode-handlers` under `with-eval-after-load 'desktop`.
- The registration lives in ghostel.el because `desktop-read` loads ghostel itself through the saved major mode and looks the handler up immediately, before anything could load ghostel-desktop.el.

Saving records `(DIRECTORY IDENTITY)`.  Restoring reuses a live buffer with the saved slot identity (so a repeated `desktop-read` doesn't spawn duplicates), otherwise starts a fresh shell in the directory under the exact saved buffer name, with the saved identity as its `ghostel-identity`. Window configurations find the buffer by name, and `ghostel-project` and friends keep reusing the restored terminal.  A failed spawn (e.g. the directory no longer exists) kills the partial buffer and re-signals so desktop.el reports the buffer and moves on.

Two kinds of saved terminals are deliberately skipped, with a message:

- **Command buffers** (`ghostel-exec`, `ghostel-compile`, eshell visual commands — identities with a `command` key).  Unlike a bookmark jump, a desktop restore runs unattended at startup, and automatically re-running a saved command (e.g. a build) seemed like a footgun.  Happy to add an opt-in defcustom for respawning such commands if you'd like parity with the bookmark jump.
- **Remote (TRAMP) terminals**, since spawning their shells would open connections, possibly prompting for credentials, in the middle of `desktop-read`.  As above, an opt-in is easy if wanted.

I kept v1 free of new defcustoms given the API overhaul you mentioned. Both skips are documented in the README's new Desktop section.

## Other info

Eleven tests modeled on ghostel-bookmark-test.el.

README section, comparison-table row, and CHANGELOG entry included.

Restoring many terminals spawns one shell per buffer. The README points at the stock `desktop-restore-eager` for throttling rather than anything ghostel-specific.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Ghostel buffers survive `desktop-save-mode` restarts.  A saved
+  terminal records its working directory and identity and
+  `desktop-read` starts a fresh shell in that directory under the
+  saved buffer name, so restored window configurations find the
+  buffer.
+
 ### Changed
 - The bookmark functions are now public: `ghostel-bookmark-make-record`
   and `ghostel-bookmark-handler`.  Bookmarks saved under the old
--- a/README.org
+++ b/README.org
@@ -726,6 +726,27 @@ fresh buffer starts its shell in the bookmarked directory, and a reused buffer
 that has since moved elsewhere gets a =cd= typed into its shell.  Set it to =nil=
 to leave the directory alone.
 
+** Desktop
+:properties:
+:custom_id: desktop
+:end:
+
+Ghostel integrates with desktop.el to restore terminal buffers on restart.  As
+with bookmarks, only the directory and identity are saved and restored, not
+scrollback or session contents.
+
+Some saved terminals are skipped at restore time (reported in =*Messages*=).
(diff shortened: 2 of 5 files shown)


dakra:
Haven't looked too closely but looks good on a quick pass.
Can you rebase on latest master and fix CI (which could be tricky with the not allowed with-eval-after load, not sure what's best there or if we have to add an exception)

mrcnski:
Thanks! I plan to update all my PRs soon but haven't been able to do so yet.

dakra:
@mrcnski thanks. I rebased, fixed the CI, the lint warnings and some other edge cases my clanker complained about (restore when the directory doesn't exists anymore, dead terminals and buffer renames)

Merged 🙏

PR #631, opened 2026-08-13
Read the default shell from the global environment

jcmfernandes:
`ghostel-shell` defaults to (getenv "SHELL"), evaluated when ghostel loads -- and ghostel is normally autoloaded by the first terminal, so that read happens in whatever buffer was current at that moment.

For example, `buffer-env`, `envrc` and `mise.el` bind `process-environment` buffer-locally to a project's dev shell, so opening the first terminal from such a buffer pinned every later terminal to that project's shell. Dev shells rarely name an interactive shell: a nix devshell exports a bash built without readline, which prints the `\[` / `\]` zero-width markers of the user's PS1 literally and turns every prompt into garbage.

Read past any buffer-local binding so the buffer that happens to open the first terminal cannot decide which shell the user gets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- The default of `ghostel-shell` reads `$SHELL` from the global
+  environment instead of the current buffer's. Ghostel is normally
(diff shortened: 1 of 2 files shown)


dakra:
Thanks.
I removed the huge clanker comment and put it directly inside the defcustom.
merged 🙏

PR #632, opened 2026-08-13
Suppress point-leave copy-mode entry on minibuffer arrivals

### LINKED ISSUE #630: Overly aggressive read-only mode
Hi, thanks for this great project.

Perhaps since a week ago I noticed that opening new terminals with (at least) `ghostel-project` suddenly turns the original buffer I was in into read-only mode, and I have to manually disable it every time. 

Here are my settings:
```elisp
(after! ghostel
  (map! :map ghostel-char-mode-map
        "<escape>" #'ghostel-semi-char-mode)
  (setq ghostel-initial-input-mode 'char)
  (add-hook 'ghostel-exit-functions #'colin/vterm-kill-window-on-exit))
```
and I have keybindings bound to this function, which is how I always invoke `ghostel`:
```elisp
;;;###autoload
(defun colin/new-ghostel-over-there ()
  "Spawn a new window to the right and open a fresh Ghostel session in it."
  (interactive)
  (+evil/window-vsplit-and-follow)
  (ghostel-project t)
  (evil-insert 0))
```
I also tested `+evil/window-vsplit-and-follow` in isolation and it is not the culprit. Any ideas? Thank you kindly.

dakra:
Fixes #630.

## Problem

Any minibuffer exit ran the point-leave check in whatever ghostel buffer occupied the selected window afterwards. A minibuffer command that merely **switches to** a ghostel buffer (`consult-buffer`, `C-x b`, `ghostel-list-buffers`) is an arrival, not navigation: the freshly displayed window can restore a stale point (a `window-prev-buffers` marker from an older layout, often the buffer end), and the deferred check saw the mismatch before the redisplay anchor could re-sync point — freezing the buffer into read-only copy mode seemingly at random. Live traces from an affected session showed both reported flavors: restored point at `point-max`, and restored point at position 1, each flipping the buffer on `vertico-exit`.

## Fix

- `ghostel--minibuffer-exit-maybe-leave` derives the minibuffer's originating buffer at exit time from `minibuffer-selected-window` (per-level correct, so nested minibuffers cannot clobber it) and only runs the deferred check when the exit lands back in that same ghostel buffer. For non-ghostel origins no timer is scheduled at all.
- Point at `point-max` counts as riding the cursor, now shared with the window-follow logic through the new `ghostel--pos-on-cursor-p`.

The intended behavior is unchanged: minibuffer navigation from within the buffer (`consult-line`, `goto-line`) still switches to the configured read-only mode, as do isearch and advised jump commands.

## Verification

- 4 new ERT tests (arrival suppressed, origin navigation fires, non-ghostel origin schedules nothing, point-max parity); full `make all` green.
- Deterministic live A/B in a sandboxed session: on `main` the arrival repros flip to `(copy t …)`; on this branch they stay `(semi-char nil …)`, while `goto-line` from inside the buffer still enters copy mode on both.

### DIFF
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -2695,9 +2695,17 @@ command set the region, so the selection survives the switch."
              (not ghostel--inhibit-insert-forwarding))
     (ghostel--enter-readonly-input-mode ghostel-mark-activation-input-mode)))
 
+(defun ghostel--pos-on-cursor-p (pos)
+  "Non-nil if POS rides the live terminal cursor: on it, or at `point-max'.
+Positions between the cursor and `point-max' do not count."
+  (and ghostel--cursor-char-pos
+       (or (= pos ghostel--cursor-char-pos)
+           (= pos (point-max)))))
+
 (defun ghostel-maybe-leave-input (&rest _)
   "Leave semi-char for `ghostel-point-leave-input-mode' if point left the input.
-A no-op unless, in semi-char mode, point has moved off the live terminal cursor.
+A no-op unless, in semi-char mode, point has moved off the live terminal
+cursor (`ghostel--pos-on-cursor-p', so `point-max' counts as on it).
 Wired into `isearch-mode-end-hook' and `minibuffer-exit-hook'.
 Add it to other jump commands as a hook or `:after' advice (see the README)."
   (interactive)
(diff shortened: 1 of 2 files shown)


fosskers:
Unfortunately this still locks the original buffer every time.

Issue #611, opened 2026-08-07
Cannot paste screenshot to terminal using `C-y` or `M-x yank-media`


pkradiator:
Basically the title.

I am running claude code inside ghostel (which is inside emacs wsl x-window) and I tried to past using `C-y` but It didn't work. I also tried `M-x yank-media` but that also didn't work. Surprisingly `C-v` worked

dakra:
This is because `C-v` is a claude-code keybinding. When pressed it reads the OS native clipboard and pastes it directly from there. Ghostel is not involved.

Basically there is no "paste image to the terminal". Not in any terminal.
We could make a workaround similar to drag'n'drop, that most TUIs that accept images (like claude-code)
take a file-path. So on `yank-media` we could write a temp file and then paste the file path.
For claude-code that would result like you pasted the image.

dakra:
In latest main I implemented `yank-media` for images and pdfs.
It copies the clipboard to a temp file and then pastes the file path, similar to what a drag-and-drop of a normal file does.

Hope the earlier explanation and the yank-media support is enough.

Closing the issue now but please reopen or create a new one if you find something else.

PR #638, opened 2026-08-16
Paste file drops and clipboard media through the paste encoder

mrcnski:
Follow-up to #626.

## Problem

#626 delivers file drops and clipboard-image pastes by *typing* the shell-quoted path: `ghostel--dnd-send-files` sends raw keystrokes, while dropped *text* already goes through the paste encoder. The running program can't tell a dropped path from typed input — which defeats the use case #611 was closed for: Claude Code attaches a dropped or pasted image as `[Image #N]` only when the path arrives inside bracketed paste. Typed, it stays plain text. (Warp has the same bug: anthropics/claude-code#48153.)

Delivering drops through the paste path is the terminal convention, including in Ghostty on both apprts ([macOS](https://github.com/ghostty-org/ghostty/blob/bedb6f2ff9d2c65456a7b1e98a5a04a702b77f27/macos/Sources/Ghostty/Surface%20View/SurfaceView_AppKit.swift#L2295-L2309), [GTK](https://github.com/ghostty-org/ghostty/blob/bedb6f2ff9d2c65456a7b1e98a5a04a702b77f27/src/apprt/gtk/class/surface.zig#L2680-L2727)) — the core [`textCallback`](https://github.com/ghostty-org/ghostty/blob/bedb6f2ff9d2c65456a7b1e98a5a04a702b77f27/src/Surface.zig#L3303-L3313) both hand the escaped paths to is documented as paste-equivalent:

```zig
/// This will treat the input text as if it was pasted
/// from the clipboard so the same logic will be applied. Namely,
/// if bracketed mode is on this will do a bracketed paste.
```

[iTerm2](https://github.com/gnachman/iTerm2/blob/3ec57866cd9bcf932f2675f7ca47183793a37b79/sources/TerminalView/PTYTextView.m#L4269-L4309), [kitty](https://github.com/kovidgoyal/kitty/blob/48625a824d2190a25533b186a35f8fd4dc641e68/kitty/window.py#L2262-L2273), and [Alacritty](https://github.com/alacritty/alacritty/blob/1b2b36a64e88068ad02c95fad00ee2fad31c00bf/alacritty/src/event.rs#L2007-L2010) all paste drops too; [VS Code's terminal](https://github.com/microsoft/vscode/blob/7314b01d35de663df9f0a0893a923f6cc16502ae/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1390-L1400) is the outlier that types.

Three smaller gaps in the keymap path:

- `ghostel--drop` ignores the event position and acts on the selected window's buffer, so dropping onto a visible-but-unselected terminal pastes into the wrong buffer or nowhere. (`ns-drag-n-drop` selects the posn window for exactly this reason.)
- Drops on the mode line vanish silently. They arrive as `[mode-line drag-n-drop]`; mode-line `local-map` text properties replace the buffer's local map during the lookup, and the area-prefixed sequence doesn't fall back to the global `[drag-n-drop]` binding, so it resolves to nothing. Easy to hit in practice — TUIs like Claude Code draw their input box right above the mode line.
- Char mode doesn't bind `<drag-n-drop>` (the map is parentless), so the drop falls through to the global handler, which steals the window selection and raises the frame.

## Fix

`ghostel--dnd-send-files` pastes instead of typing, keeping the shell-quoting and the trailing space. All three delivery routes share the helper, so X11/pgtk drops, `yank-media` pastes, and keymap drops are fixed at once. The encoder brackets only when the child enabled mode 2004; other programs receive byte-for-byte what they got before.

`ghostel--drop` runs in the buffer of the event posn's window — the terminal actually dropped on. The posn is read as `(nth 1 event)` rather than via `event-start`, whose `posn-at-point` fallback would invent a selected-window position for a posn-less event.

A new `ghostel--dnd-area-map` binds `[AREA drag-n-drop]` for the mode line, header line, and fringes, registered in the existing emulation alist: emulation maps rank above `local-map` text properties, which is the only way those sequences are reachable. `ghostel-char-mode-map` binds `<drag-n-drop>` directly.

## Other info

test/ghostel-dnd-test.el grows from 8 to 16 tests. The existing dispatch tests only change their capture point (`ghostel--paste-text` instead of `ghostel--send-string` — every assertion unchanged). New: six unit tests for `ghostel--drop` (quoting and joining, posn targeting, keymap and decoration-area coverage) and two native tests asserting the exact PTY bytes with and without mode 2004.

Verified live against Claude Code: a dropped screenshot attaches as `[Image #N]`; the same path typed stays plain text. The mode-line claim was checked with `key-binding`'s POSITION argument on a real mode-line posn, where the text-property override applies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,15 @@ All notable changes to this project will be documented in this file.
   ran. A nix devshell exports a `$SHELL` built without readline, which
   prints the `\[` / `\]` zero-width markers of the user's prompt
   literally.
+- Dropped files now go through the paste encoder like dropped text
+  does, so a program that enabled bracketed paste sees the
+  shell-quoted path arrive as a paste — matching other terminals.
+  TUIs that special-case pasted paths (e.g. Claude Code attaching a
+  dragged image) recognize drops again.
+- File and text drops now land in the terminal shown in the window
+  under the pointer; previously the drop acted on the selected window's
+  buffer, so dropping onto an unselected terminal window went to the
+  wrong buffer or nowhere.  Char mode accepts drag-and-drop now as well.
 
 ## [0.50.0] — 2026-08-13
 
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -1176,8 +1176,18 @@ is non-nil."
   "<wheel-up>"   #'ghostel--scroll-intercept-up
   "<wheel-down>" #'ghostel--scroll-intercept-down)
 
+(defvar-keymap ghostel--dnd-area-map
+  :doc "Keymap for drops on a terminal window's decorations.
+Mode-line `local-map' text properties shadow the buffer's local map,
+so these bindings must live in an emulation map."
+  "<mode-line> <drag-n-drop>"    #'ghostel--drop
+  "<header-line> <drag-n-drop>"  #'ghostel--drop
+  "<left-fringe> <drag-n-drop>"  #'ghostel--drop
+  "<right-fringe> <drag-n-drop>" #'ghostel--drop)
+
 (defvar ghostel--emulation-alist
-  `((ghostel--scroll-intercept-active . ,ghostel--scroll-intercept-map))
+  `((ghostel--scroll-intercept-active . ,ghostel--scroll-intercept-map)
+    (ghostel--term . ,ghostel--dnd-area-map))
   "Alist for `emulation-mode-map-alists'.")
 
 (unless (memq 'ghostel--emulation-alist emulation-mode-map-alists)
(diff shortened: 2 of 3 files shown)


dakra:
Thanks, merged (with some minor changes).

PS, you really have to reign in Fable and Opus 5 regarding code comment as they tend to go pretty wild :D
(see https://github.com/dakra/ghostel/compare/343b48601cc8c727d834b8192ed2961474187b5b..ff3d83e7f60326c93c22a1ef23aa4cc22907374d#diff-ed8e7ac1f2d2973df1e0703d9c95b7142e3d0c4dc2330f1c45a5500f7cf9eaa5 )

mrcnski:
Thanks for the cleanup, and noted your feedback. I was in a rush when I made this PR and I admit it was a bit sloppy. Thanks!

dakra:
No worries. Was also not really meant as negative feedback or anything.
More my personal complain about the verbosity of Fable/Opus output :D

PR #639, opened 2026-08-16
Fix stale titles after RIS and implement the xterm title stack

dakra:
Fixes the two stale-title paths reported in the [follow-up on #619](https://github.com/dakra/ghostel/issues/619#issuecomment-5280801420), plus one adjacent case of the same bug class.

## RIS (`ESC c`) never reached elisp

`Terminal.fullReset()` clears the stored title, but the `full_reset` stream arm never fires the `title_changed` effect, so `ghostel--title` (mode line, title-derived buffer names) kept the pre-reset title after `reset`(1). The handler now intercepts `full_reset`, forwards it, and fires the title callback — gated on a title actually having been set, so a bare `reset` stays free of elisp side effects.

Same class, same interception point: a program reporting ConEmu OSC 9;4 progress that gets `reset` left the (default-on) progress indicator stuck. The arm now dispatches a progress `remove`, mirroring ghostty's own `StreamHandler.fullReset`.

## `CSI 22/23 t` window-title stack

Previously in the no-effect arm, so a program that saves the title, sets its own, and restores it on exit left its final title stuck. This is not theoretical: vim's xterm detection is a TERM name-prefix match that `xterm-ghostty` satisfies, so vim pushes/pops the stack unconditionally, and with `set title` it left "Thanks for flying Vim" in the mode line on every exit. (Ghostty proper has the same gap but masks it by rewriting the title from shell integration at every prompt; ghostel's shell integration deliberately never touches the title, so nothing self-heals.)

Implementation, kept minimal:
- bounded per-terminal stack (10 entries, xterm's documented limit); pushes beyond the limit are dropped
- only the window-title selectors reach the handler (the parser drops icon-only forms), and only the default stack index 0 acts — slot addressing stays unimplemented
- pop restores through the standard `window_title` path so native terminal state and the elisp callback stay in sync; restoring a no-title push clears the title
- RIS and handler deinit discard the stack

## Verification

- 7 new native ERT tests (both PTY backends), including a byte-for-byte replay of vim's captured push/set/pop sequence, a pop→push round trip that pins native/elisp sync, an 11-push overflow test, and a FIFO-sentinel test proving the RIS gate
- red/green checked: the RIS tests fail against the unfixed handler, and the gate test fails against an ungated one
- live-verified in a real session: vim exit leaves no stale title; `reset` clears a stale title and a stale progress dispatch
- `make -j8 all` green

### DIFF
--- a/src/GhostelTerm.zig
+++ b/src/GhostelTerm.zig
@@ -54,7 +54,7 @@ pub fn init(
     };
     errdefer term.terminal.deinit(alloc);
 
-    term.stream = .initAlloc(alloc, .init(term, &term.terminal));
+    term.stream = .initAlloc(alloc, .init(alloc, term, &term.terminal));
     errdefer term.stream.deinit();
 
     term.renderer = try .init(alloc, env, &term.terminal);
--- a/src/NativeProcess.zig
+++ b/src/NativeProcess.zig
@@ -70,7 +70,7 @@ pub fn init(
     var event_writer = try EventWriter.init(event_fd);
     errdefer event_writer.close();
 
-    var stream: @TypeOf(self.stream) = .initAlloc(alloc, .init(self, term));
+    var stream: @TypeOf(self.stream) = .initAlloc(alloc, .init(alloc, self, term));
     errdefer stream.deinit();
 
     const replica_name = try alloc.dupeZ(u8, backend.replicaName());
--- a/src/handler.zig
+++ b/src/handler.zig
@@ -2,19 +2,37 @@
 //! standard terminal handler but intercepts OSC-related actions so we can route
 //! them to Elisp callbacks instead of re-parsing the same bytes ourselves.
 
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
 const emacs = @import("emacs.zig");
 const gt = @import("ghostty-vt");
 const GhostelTerm = @import("GhostelTerm.zig");
+
+const log = std.log.scoped(.GhostelHandler);
+
+// xterm limits the title stack to 10 entries.
+const title_stack_max = 10;
 
 pub fn GhostelHandler(Context: type) type {
     return struct {
         const Self = @This();
 
+        alloc: Allocator,
         context: Context,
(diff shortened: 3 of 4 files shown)


dakra:
@emil-e the first 2 commits are simple
the last one with the title stacks is something that Ghostty doesn't even have implemented, but there nobody really noticed I guess because they change the title in their shell injection in every prompt.
We don't do that, so when you're in e.g. bash, start vim, `:set title` and leave vim, you're left with a title of 'Thanks for flying Vim'.

mrcnski:
Thanks for picking this up so quickly — and the vim case makes the stack half far less theoretical than my report did. Verified locally on macOS aarch64 (Emacs 30.2, ReleaseFast module): built this branch and ran the native OSC suite, 26/26 on both pty backends. The index-0 gate also matches what the pinned parser actually forwards for `CSI 22/23 t` (icon-only selectors filtered out, payload is the stack index), so the indexed forms correctly no-op.

One stale-title path from my #619 follow-up remains after this lands: programs that never ask for cleanup at all — no RIS, no title stack, no clear on exit. Claude Code is the concrete case: it sets `ESC ] 0 ;` session summaries while it runs, but (checked against the installed binary) it ships no `CSI 22/23 t` and no restore-on-exit, so every quit still leaves the last summary sitting next to an idle prompt. As this PR's description notes, nothing self-heals here because ghostel's shell integration deliberately never rewrites the title.

I have a pure-elisp fix for that class working locally: with shell integration active, titles become command-scoped — the OSC 133 C marker saves the current title, and D reverts to it if the command changed it, after the finish hooks (which still see the command's title). Bare D markers from prompt redraws leave prompt-set titles alone, and programs that clean up properly are unaffected: by D time a stack pop or an explicit clear has already restored things, so there is nothing to revert. New defcustom `ghostel-restore-title-after-command` (default t), no module changes.

Happy to raise it as a follow-up PR rebased on this branch once it lands — its tests slot into the same region of test/ghostel-osc-test.el as the new ones here.

dakra:
merged in main

PR #641, opened 2026-08-16
Add a public terminal-buffer API and the consult-ghostel extension

dakra:
Two commits:

**Add a public terminal-buffer API** (`lisp/ghostel.el`)
- New `ghostel-create`: create a fresh terminal running `ghostel-shell`, with optional name, display action, and identity override; always creates (reuse is the caller's job). `ghostel--start` now routes through it.
- `ghostel--all-buffers` → `ghostel-buffer-list`, `ghostel--project-buffers` → `ghostel-project-buffer-list` (public, no aliases).
- README section on creating and listing terminals from Lisp; tests for slot-identity allocation, verbatim identity override, and empty-name handling.

**Add the consult-ghostel extension package** (`extensions/consult-ghostel/`)
- Independent MELPA package (like evil-ghostel): `consult-ghostel` / `consult-ghostel-project` pick a terminal via `consult--multi` with live preview, MRU-first ordering, and create-on-miss through the new public API (project variant stamps a project-rooted identity). Hidden source variants for `consult-buffer` / `consult-project-buffer`.
- Candidates are annotated with the terminal title via `ghostel-annotate-buffer`; with marginalia installed, the title is prepended to marginalia's buffer annotations instead through its annotator registry (works in every buffer prompt, no configuration).
- Loading the package makes `consult-line` match across soft line wraps in ghostel buffers, with wrap-corrected point placement.
- Makefile/CI: consult + marginalia checkouts for compile and tests, `test-consult` target and CI job, melpazoid job for the new recipe.

Note: the `(ghostel "0.50.0")` dependency floor must be bumped to 0.51.0 at the release that first ships `ghostel-create`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,42 @@ jobs:
             -l test/evil-ghostel-test.el \
             -f evil-ghostel-test-run
 
+  test-consult:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        emacs_version:
+          - '29.4'
+          - 'snapshot'
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v31
+        with:
+          github_access_token: ${{ github.token }}
+
+      - name: Install pinned Emacs
+        env:
+          EMACS_VERSION: ${{ matrix.emacs_version }}
+        run: bash .github/scripts/install-emacs.sh
+
+      - name: Install consult
+        run: |
+          git clone --depth 1 https://github.com/minad/consult.git /tmp/consult
+
+      - name: Install marginalia
+        run: git clone --depth 1 https://github.com/minad/marginalia.git /tmp/marginalia
+
+      - name: Install compat
+        run: git clone --depth 1 https://github.com/emacs-compat/compat.git /tmp/compat
+
+      - name: Run consult-ghostel tests
+        run: |
+          emacs --batch -Q -L /tmp/compat -L /tmp/consult -L /tmp/marginalia -L lisp -L extensions/consult-ghostel \
+            -l ert \
+            -l test/consult-ghostel-test.el \
+            -f consult-ghostel-test-run
+
   build-native:
     strategy:
       fail-fast: false
--- a/.github/workflows/melpazoid.yml
+++ b/.github/workflows/melpazoid.yml
@@ -67,3 +67,30 @@ jobs:
           EXIST_OK: false
           WARN_IS_ERROR: true
         run: make -C ~/melpazoid
+
+  build-consult:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with: { python-version: '3.10' }
+      - name: Install
+        run: |
+          sudo apt-get install emacs && emacs --version
+          git clone https://github.com/riscy/melpazoid.git ~/melpazoid
+      - name: Install dependencies
+        run: |
+          emacs --batch -Q \
+            --eval "(require 'package)" \
+            --eval "(add-to-list 'package-archives '(\"melpa\" . \"https://melpa.org/packages/\") t)" \
+            --eval "(package-initialize)" \
+            --eval "(package-refresh-contents)" \
+            --eval "(package-install 'consult)"
+      - name: Run
+        env:
+          LOCAL_REPO: ${{ github.workspace }}
+          RECIPE: (consult-ghostel :fetcher github :repo "dakra/ghostel" :files ("extensions/consult-ghostel/consult-ghostel.el"))
+          EXIST_OK: false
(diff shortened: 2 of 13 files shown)


mrcnski:
Nice.  Once this is merged, I'll update my personal dwim command (#620) to hand
off to `consult-ghostel-project` for live previews.  But at that point, I'm not
sure if my dwim command should be included in ghostel core (and you had already
expressed hesitation), and now the case gets even weaker.  Instead of supporting
multiple dwim shapes (consult vs. non-consult) in core, I can add my command to
the readme (or some other appropriate reference).

jessy836:
A nice integration to have with consult would be the ability to use consult-history.

jessy836:
An other one, would be consult-bookmark-narrow to be able to narrow bookmark to keep only ghostel's bookmarks

mrcnski:
One nit from the AI:

Bookmarks saved before the 0.51.0 rename carry `ghostel--bookmark-handler` in their records — the obsolete alias keeps jumping working, but records never re-stamp, and consult resolves the narrow group by the stored symbol, so every pre-rename bookmark lands in "Other" instead of the Ghostel group. Listing both symbols fixes it, same as consult's own multi-handler entries:

```elisp
(push '(?g "Ghostel" ghostel-bookmark-handler ghostel--bookmark-handler)
      consult-bookmark-narrow)

jessy836:
Thank you very much for implementing both suggestions so quickly.

I see that you created the new `consult-ghostel-history` command. I was wondering whether, for consistency with the rest of Consult, it could also be useful to make the existing `consult-history` command work in Ghostel buffers.

It could be great to have both:

- `consult-history` as the familiar, basic history interface, so users can keep the same command and habits across Eshell, Comint, term, and Ghostel;
- `consult-ghostel-history` as the richer Ghostel-specific command.

One possible approach for `consult-history` might be to use a `:before` advice to populate a history variable from `ghostel-shell-history` when the command is invoked in a Ghostel buffer, and let `consult-history` use that variable. I am not sure whether this would be the cleanest implementation, but perhaps it could be a useful direction.

For `consult-ghostel-history`, I also wonder whether it could reuse Consult's asynchronous facilities. The Consult interface could open immediately while the real shell history is retrieved in the background, with candidates appearing progressively. This could be particularly useful for remote terminals, large histories, or history managers such as Atuin.

Having both the standard `consult-history` integration and the more advanced asynchronous `consult-ghostel-history` command could be a really nice combination. What do you think?

Full disclosure: I discussed these ideas with Codex GPT-5.6 Terra to research the relevant Consult and Ghostel APIs and to help draft this comment.

dakra:
> Bookmarks saved before the 0.51.0 rename carry ghostel--bookmark-handler in their records

I know and purposely removed that extra code. I don't think many users have existing bookmarks anyway and instead of always having to add extra code for maybe 1 or 2 users, I don't support consult-bookmark for them in the hopes they just create new bookmarks.
I also plan to remove the defalias soon.

dakra:
@jessy836 thanks for the comments.

> It could be great to have both:

Can you expand what you expect from `consult-history` and why `consult-ghostel-history` is not a appropriate replacement? i.e. what would the difference be between consult-history and consult-ghostel-history?

In the Ghostel version we send the text to the PTY (and press ctrl-c before) which normal consult-history (obviously) doesn't do.

> For consult-ghostel-history, I also wonder whether it could reuse Consult's asynchronous facilities. The Consult interface could open immediately while the real shell history is retrieved in the background

I don't think that makes much sense. Since it's not like a bigger grep search or something where you get to see the first 20 candidates and then more load in the background.
Popping up a empty history buffer is imho more confusing than helping. And it's not like the 100 to 1000 history items come in incrementally, they're either all there or not.
But maybe you have a different use-case or other history tools?

jessy836:
@dakra Thanks, that makes sense.

TBH, `consult-ghostel-history` already does the job very well, and it is more capable than the regular `consult-history`. My suggestion to support both is mostly nitpicking rather than something I actually need.

The main advantage I see in also supporting `consult-history` would be workflow consistency across multiple Emacs terminal and REPL environments. A user could bind `consult-history` once and expect that binding to work in Eshell, Comint, term, Ghostel, and other supported modes. The benefit would therefore be familiarity and uniformity, not functionality missing from `consult-ghostel-history`.

Regarding asynchronous retrieval, I do not currently have a concrete use case that requires it. My thought came mostly from the architecture introduced by this PR: Ghostel now provides the dedicated `ghostel-shell-history` API, and `consult-ghostel` already builds the Consult integration around it. This seemed like much of the foundation needed to eventually connect the history retrieval to Consult's asynchronous facilities was now in place, so I thought it could be a nice way to make the integration as complete and advanced as possible.

However, I understand your point that shell history is generally small, is normally retrieved all at once, and that opening an empty picker while waiting may be more confusing than useful. So this is also mostly a refinement idea, not something I consider necessary or something that should block the PR.

The current `consult-ghostel-history` implementation already covers my actual needs very well. Thank you for implementing it.

PR #648, opened 2026-08-23
Use Fish shell's default $hostname env var

dlozeve:
Some systems do not have a `hostname` command, and Fish has an environment variable for this:

https://fishshell.com/docs/current/language.html#envvar-hostname


dakra:
Thanks 🙏

PR #655, opened 2026-08-26
Improve evil experience

ivanrg99:
Hi! This project is amazing. It feels rock solid and is incredibly fast, great work!

What held me back from using it more was evil support. Whenever I would use vim motions like `ciw`, or just delete things around, the cursor/point would feel "stuck", like it would sometimes not delete a character and leave some garbage in the ghostel buffer. It would also not understand my evil motions fully.

I have fixed evil support so that it handles wide glyphs and graphemes correctly, and updated the syntax table so that it "understands" vim motions. I am happy to report that I have found all my issues with ghostel fixed with these changes! The evil experience is now truly silky smooth and I have dropped vterm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ All notable changes to this project will be documented in this file.
   coupling correctness to redraw scheduling.  Incremental redraws skip clean
   rows above the first changed row, reducing the cost of the per-key feedback
   loop.
+- `evil-ghostel`: `C-c C-t` now enters copy mode in Evil normal state and
+  printable Vim navigation keys no longer trigger Ghostel's read-only fast
+  exit.  Copy mode stays frozen until `C-c C-t` or another explicit
+  input-mode switch exits it; accidentally entering insert state remains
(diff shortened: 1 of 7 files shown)


ivanrg99:
I have also fixed `copy-mode` so that it works well with `evil`, plus some minor performance optimizations in the rendering

dakra:
@ivanrg99 thanks for the PR and the nice words.

I had a look when you submitted your first commit and make a few changes to the word boundary parts that is now on main with PR #661 

Is the PR finished otherwise or are more commits coming? I'll have a look at the rest then.

ivanrg99:
@dakra Yes, the PR should be ready for review. I removed the code in the commits that were related to the word boundary since those are no longer needed

dakra:
Great thanks. Is it possible for you to open a PR for each feature separately?

So I think I prefer 3 PRs instead of 1 that mixes 3 (previously 4) different things.
Makes it easier to apply the simper changes (copy-mode) while discussing the more difficult ones.

Thanks ❤️

ivanrg99:
Sure, no worries! I'll close this PR and create new ones

PR #657, opened 2026-08-28
Report terminal color scheme via CSI 996 and Mode 2031

ayanamists:
## Summary

- Answer CSI `? 996 n` with the terminal's dark/light color scheme.
- Emit CSI `? 997` after synchronized default colors change when Mode 2031 is enabled.
- Classify the last synchronized `ghostel-default` background using Neovim's BT.601 midpoint convention.
- Keep child OSC 11 overrides separate from the host color scheme.

## Why

libghostty already parses CSI 996 and tracks Mode 2031, but Ghostel did not provide the color-scheme callback or send change reports. Multiplexers could therefore keep stale host color caches after an Emacs theme change.

Ghostel does not store a separate light/dark appearance value. Its protocol background is seeded from the resolved `ghostel-default` face, which users can customize independently from the rest of Emacs. CSI 996/997 therefore classify the most recently synchronized default background. Theme changes synchronize it through the existing hook; after changing `ghostel-default` directly, users can call `ghostel-sync-theme`.

The implementation deliberately reads the default background rather than the current OSC 11 override. An OSC 11 override belongs to the child-controlled terminal session and should not redefine the host terminal's appearance.

The RGB classification follows the same BT.601 luma and 0.5 midpoint convention used by [Neovim's terminal background detection](https://github.com/neovim/neovim/blob/2fbc82820b5168b0da4e918c192c05832dfd6211/runtime/lua/vim/_core/defaults.lua).

## Testing

- `zig build --prefix .`
- `zig build test`
- Native OSC test suite: 30/30 passed
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 - Color-scheme DSR is implemented: CSI `? 996 n` reports light/dark
-  from the protocol background (OSC 11), and enabling Mode 2031 now
-  actually produces CSI `? 997` reports when the Emacs theme changes.
+  from the last synchronized `ghostel-default` background, and enabling
+  Mode 2031 now actually produces CSI `? 997` reports when the Emacs theme
+  changes or `ghostel-sync-theme` is called explicitly.
   Previously ghostel stored Mode 2031 but never answered 996 or pushed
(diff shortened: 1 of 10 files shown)


dakra:
@ayanamists thanks.
I squashed your commits and merged them as part of PR #659 

I didn't make a release yet, so if you want to use this feature you have to build the native module yourself until a release is made.

Issue #97, opened 2026-04-13
Mouse scrolling is very fast


affe-mit-waffel:
Hi there 👋,

I scroll through buffers using a mouse / trackpad on a regular basis. Scrolling inside a ghostel buffer is very sensitive.
The cause seems to be, that `ghostel--scroll-up` does call `(scroll-down 3)` and `ghostel--scroll-down` does call `(scroll-up 3)`. Is that a deliberate choice?

Here two videos showing scrolling through some text (output of `brew list` twice). I'm using a MacBook trackpad (very carefully). It's worse on my Logitech Mouse.

https://github.com/user-attachments/assets/81e65177-4040-4085-b253-4e613f8f41da

After unmapping `<wheel-up>` and `<wheel-down>` for testing:

https://github.com/user-attachments/assets/b248f576-9397-4bb0-9f62-f30da7761075

 I use [ultra-scroll](https://github.com/jdtsmith/ultra-scroll).

Possible solutions:

- Allow to user to customize the "scroll factor" instead of the hardcoded `3`.
- Fallback to the scrolling the user has setup for themself. This way e.g. `pixel-scroll-precision-mode` and `ultra-scroll` would be supported as well. I'm not sure how to achieve this though (but shouldn't be too complicated?)

dakra:
I disabled it in https://github.com/dakra/ghostel/commit/067af2592d520e1ed3d39d224dc6ead4162077e7 because ultra-scroll and pixel-scroll-precision-mode hijacked the key(/mouse)-binding.
But since then the ghostel scrollback buffer is a normal Emacs buffer so I don't think we need the custom scroll commands anymore. I'll investigate.

dakra:
This was a bit trickier than expected as we still need to catch the mouse/touchpad events first and forward them to the tui app in the terminal when they want it (like scrolling in vim or opencode).
I published a fix in the latest version. Please check and re-open if there is anything not working.
Thanks.

affe-mit-waffel:
Yes, now it works as expected. Thank you for the quick fix! Ghostel has become such an awesome package 🔥

dakra:
Thanks ❤️

PR #660, opened 2026-08-30
Accumulate high-resolution wheel events forwarded to mouse-tracking apps

mrcnski:
Fixes wheel scrolling being far too fast in mouse-tracking apps (vim, htop, Claude Code) when a pixel-precision scroll package is active.

**Problem:** `pixel-scroll-precision-mode` and ultra-scroll disable wheel-event coalescing, so one trackpad gesture arrives as dozens of small-delta wheel events. The scroll intercept forwards each one to the terminal as a full wheel click. Follow-up to #97, which made non-tracking buffers fall through to the user's scroll package but left the forwarded path at one click per event.

**Fix:** accumulate pixel deltas per buffer and forward one wheel click per cell height of travel, as ghostty itself does. Events without pixel data (classic one-notch-per-event wheels) still forward one click each. A new `ghostel-mouse-scroll-multiplier` option (named after ghostty's `mouse-scroll-multiplier`) tunes the rate.

Since events absorbed into the accumulator must be consumed without sending anything, the tracking check moved out of the send: a new native `ghostel--mouse-tracking` exposes the terminal's `flags.mouse_event` — the same state the mouse encoder consults — and X10-only tracking falls through to the user's scroll package, since X10 never reports wheel buttons.

**Tested** on macOS (NS build) with a trackpad against Claude Code; plain-shell buffers still fall through to ultra-scroll as before. Untested on the emacs-mac port, where wheel events may lack the pixel-delta slot — the guard makes that case behave exactly as before this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ All notable changes to this project will be documented in this file.
   `bottom` moves it all above the grid, `top` (the default) keeps it
   below.  Graphical frames on Emacs 29+ only.
 
+### Changed
+- Trackpad scrolling in mouse-tracking programs (htop, vim) under
+  `pixel-scroll-precision-mode` or ultra-scroll no longer sends one
+  wheel press per trackpad tick: pixel deltas are accumulated and one
+  press is sent per row of travel, as ghostty does.  A mouse-wheel notch
+  stays one press.
+
 ## [0.52.0] — 2026-08-31
 
 ### Added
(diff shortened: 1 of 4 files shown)


mrcnski:
Looks like tests need to be updated (and maybe some new ones added). I'll try to do this a bit later.

dakra:
Thanks. I squashed the commit and made a few changes. Also removed the defcustom.
I'm not against adding one but would rather only do when someone asks for it and the emacs `nx/x-scroll-event-delta-factor` settings are not good enough.

PR #664, opened 2026-08-31
Add ghostel-window-padding-balance to align the grid vertically

### LINKED ISSUE #663: Extra vertical gap above the modeline when using tmux
Ghostel is amazing! But I found what seems to be an incorrect vertical alignment when running tmux.

In a raw Bash shell, there is no gap between the modeline and the prompt. See in the example below how the number 4 is truncated so that the prompt lands perfectly above the modeline. I think this is the proper behavior.
Image

But if I run a tmux instance, a gap appears between the tmux status bar and the modeline. See how the 6 is fully shown. The first line (6, here) is never truncated, because of the spurious gap between the tmux status bar and the modeline. I assume this is not expected, and that the proper behavior would be to remove the gap between the tmux status bar and the modeline, and show a truncated 5.
Image

The analysis of an AI is the following:

> With these measurements:
>
>     (window-screen-lines) => 73.71428571428571
>
> Ghostel normally rounds this down to 73 terminal rows. With a regular shell, `ghostel--pixel-anchor` can use the scrollback and applies a small vertical scroll, so the bottom is aligned correctly.
>
> With tmux, Ghostel is on the alternate screen, which has no Ghostel scrollback above the visible terminal. `ghostel--pixel-anchor` then returns:
>
>     (1 . 0)
>
> If Ghostel uses 74 rows (`ceiling` instead of `floor`), it returns:
>
>     (1 . 4)
>
> and the display is aligned as expected: the first line is slightly truncated and the tmux status bar is flush with the modeline.

If that helps, here is a fix I applied in my Emacs configuration, where `ceiling` makes all the difference:

```elisp
(defun my/ghostel-window-size (process windows)
  (let ((size (window-adjust-process-window-size-smallest process windows)))
    (if (and windows
             (buffer-live-p (process-buffer process))
             (with-current-buffer (process-buffer process)
               (derived-mode-p 'ghostel-mode)))
        (cons
         (car size)
         (max
          1
          (ceiling
           (apply
            #'min
            (mapcar
             (lambda (window)
               (with-selected-window window
                 (window-screen-lines)))
             windows)))))
      size)))

(setq-default window-adjust-process-window-size-function
              #'my/ghostel-window-size)
```

Thanks for the excellent work on this very useful extension!

dakra:
Fixes #663.

A window is rarely a whole number of terminal rows tall. When nothing precedes the first row (alternate screen, unscrolled primary screen) the leftover pixels sit as a gap below the last row — so tmux's status bar floats above the mode line, while a fixed top row (htop, `emacs -nw`) looks truncated.

## What this adds

`ghostel-window-padding-balance` (graphical frames on Emacs 29+ only) picks where the leftover goes:

- `top` (default) — below the last row, as before.
- `center` — split between top and bottom, like ghostty's `window-padding-balance`. `t` is accepted as `center`.
- `bottom` — all of it above the first row, so the last row sits flush against the mode line. Natural for a shell under tmux, and consistent with how the primary screen already clips at the top once scrollback exists.

Any other value falls back to `top`.

## Implementation

The pad is a `before-string` display line above grid row 1 — an empty overlay at `point-min` whose before-string is a tiny-face newline carrying `line-height` = pad pixels. A `line-height` on row 1's *own* newline was tried first but redisplay drops it when the row fills the window width (`emacs -nw` menu bar, full echo lines), and the native renderer rewrites that newline; the separate pad line is immune to both. The pad is buffer-wide (the smallest graphical window showing the buffer bounds it), removed once scrollback precedes row 1, and dropped on major-mode change and buffer reuse.

A leftover of a row or more never pads: it means the grid does not fill the window yet (startup, pre-resize), or the body-capped pixel measurement lost the pad row. Without this guard, enabling the option before the terminal started seeded an oversized pad that ratcheted up without bound (found in live testing); such a pad now self-heals on the next anchor.

## Default = no impact

With `top` the pad overlay is never created and `window-start`/`vscroll` are byte-for-byte what they were before; the anchor path adds ~0.28 µs of pure elisp per anchor (no extra `window-text-pixel-size` call), well under 1% of an anchor.

## Testing

- 11 native ERT tests (`test/ghostel-scroll-test.el`): alt-screen split, floor rounding, bottom takes the full leftover, unfilled grid gets no pad, oversized-pad recovery, row-1 rewrite (full-width and empty), sub-row shrink rebalance, scrollback yield, buffer-wide min across windows, option toggling (including `t` = center), major-mode drop. The mocked pixel measurement models the real body-capped scan: a pad of a row or more is not counted.
- Verified live under GUI elate sessions: clean start with `bottom` set before the terminal (the previously wedging case) settles at the exact sub-row leftover and stays put across output floods; runtime toggling between all three values; pixelwise resize re-settling; plus the original alt-screen matrix (tmux, htop, less, vim, `emacs -nw`) and dynamic-geometry/lifecycle paths.

`make -j8 all` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- `ghostel-window-padding-balance`: places the fractional-row space a
+  window leaves under the grid whenever nothing precedes the first row
+  (alternate screen, fresh primary screen): `center` splits it between
+  the top and the bottom like ghostty's option of the same name,
+  `bottom` moves it all above the grid, `top` (the default) keeps it
+  below.  Graphical frames on Emacs 29+ only.
+
 ## [0.52.0] — 2026-08-31
 
 ### Added
--- a/README.org
+++ b/README.org
@@ -869,6 +869,12 @@ TRAMP-aware =auth-source-pick-first-password= example.
   reader and Emacs is notified only when callbacks or redraws are needed.
 - Cursor position updates even without cell changes.
 - Theme-aware color palette (syncs with the Emacs theme via =ghostel-sync-theme=).
+- =ghostel-window-padding-balance= places the fractional-row space under
+  the grid whenever nothing precedes the first row, e.g. on the alternate
+  screen: =center= splits it between the top and the bottom (as ghostty's
+  option of the same name does), =bottom= moves it all above the grid so
+  the last row sits flush against the mode line.  =top= (the default)
+  keeps it at the bottom.
 
 ** Inline images (Kitty graphics protocol)
 :properties:
--- a/lisp/ghostel.el
+++ b/lisp/ghostel.el
@@ -646,6 +646,15 @@ Customize the faces `ghostel-fake-cursor' and
 `ghostel-fake-cursor-box' to tune the appearance."
   :type 'boolean)
 
+(defcustom ghostel-window-padding-balance 'top
+  "Where a GUI window's fractional-row leftover pixels go.
+`top' keeps them below the last row, `center' (or t) splits them top
+and bottom (like ghostty's `window-padding-balance'),
+`bottom' moves them all above the first row."
+  :type '(choice (const :tag "Top (default)" top)
+                 (const :tag "Center" center)
+                 (const :tag "Bottom" bottom)))
+
 (defcustom ghostel-initial-input-mode 'semi-char
   "Input mode a freshly started `ghostel' terminal begins in.
 One of `semi-char' (default), `char', or `line'.  `line' engages on
@@ -3798,14 +3807,15 @@ kernel hostname drifts between NAME and NAME.local with network state,
 while the function `system-name' keeps the value from Emacs startup."
   (or (null host)
       (string= host "")
-      (let ((host (if (string-suffix-p ".local" host t)
-                      (substring host 0 (- (length host) (length ".local")))
-                    host)))
-        (or (eq t (compare-strings host nil nil "localhost" nil nil t))
-            (eq t (compare-strings host nil nil (system-name) nil nil t))
-            (eq t (compare-strings
-                   host nil nil
-                   (car (split-string (system-name) "\\.")) nil nil t))))))
+      (save-match-data
+        (let ((host (if (string-suffix-p ".local" host t)
+                        (substring host 0 (- (length host) (length ".local")))
+                      host)))
+          (or (eq t (compare-strings host nil nil "localhost" nil nil t))
+              (eq t (compare-strings host nil nil (system-name) nil nil t))
+              (eq t (compare-strings
+                     host nil nil
+                     (car (split-string (system-name) "\\.")) nil nil t)))))))
 
 (defun ghostel--resolve-local-executable (program)
   "Return the absolute local executable path for PROGRAM.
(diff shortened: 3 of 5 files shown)


mallet:
I have tried it. It does center a tmux instance instead of aligning it at the top. Would it be possible to change the option to allow either top alignment (current behavior), centering (to mirror Ghostty), or more importantly bottom alignment? When using Ghostel for a Bash shell, I think bottom alignment is more natural, and is actually Ghostel's behavior outside an alternate screen. Therefore, when using tmux to host a shell, it would make sense to align at the bottom too. At least it would be very helpful to have the option.

In Ghostty, there is no visual anchor like Emacs's modeline, so I understand the option was not proposed. The situation in Ghostel is a bit different, especially when switching between Emacs buffers, or with side-by-side buffers.

I have noticed a slight issue, which is extremely minor and does not really need to be fixed. I am still reporting it in case it is more important than I anticipate. With a colored prompt like
PS1="\[\e[1;38;5;24;48;5;153m\]prompt_line$\[\e[0m\] "
the top gap includes a few pixels of the same background color (on the left). See the screenshot.


Thank you again for all the work!

dakra:
@mallet can you check again, I added a bottom option so you can choose between top (default), center and bottom alignment. The color leak issue should also be fixed.

mallet:
@dakra Thank you very much. This is an excellent design, and it will be very useful.

However, at the moment, the options center & bottom both are unstable, behave exactly the same and introduce the same gaps.

In a new Bash shell (Ghostel just launched), pressing any key makes the shell jump up or down (alternately). See below the difference after pressing backspace (once or twice):


Once one command has been executed in the shell, everything is back to normal and stable.

In tmux, even after a command has been executed (inside tmux and in the hosting shell too), pressing a key makes the tmux area toggle between bottom alignment and top alignment (regardless of the option: center or bottom--but top is fine).


To be fully accurate, the very first keypress makes the tmux area move both up and down, in one go. The following keypresses make it move up or down, alternatively.

dakra:
@mallet can you try again? and what Emacs version are you running btw? Because I'm on latest master 32 and there it worked but apparently this part changed multiple times in Emacs 30 and 31.

mallet:
@dakra This is much better. Very stable, no glitches.

I have noticed a slight problem at the start of Ghostel, hence for a raw Bash shell which is not yet full of lines. In bottom mode, it is perfect. But in center & top mode, gaps are applied as if the Bash shell were an alt screen.

In top mode, the initial shell prompt is aligned at the top:

When a third line is displayed (the buffer/window is then full), it is corrected: the prompt is perfectly aligned at the bottom and the top line is truncated.

In center mode, there is the slight gap that would be applied with an alt screen:

Again, this returns to normal when the buffer/window is full.

The rest is perfect. I have tested center, bottom and top. Except at startup with center & top, I see no issue in raw Bash or tmux, even when resizing.

My exact version is GNU Emacs 30.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.50, cairo version 1.18.4) of 2025-11-22, modified by Debian.

dakra:
@mallet thanks for testing so extensively.
I merged this PR now as is. The bottom/center alignment on a fresh terminal is on purpose.
The alignment happens not if it's in primary or alt screen but if there is scrollback or not.
So in a fresh terminal, when there is no scrollback it's always top aligned (just like when you open a new Emacs buffer).
I guess we could discuss if that behavior should also change but since it's an existing behavior already on main I think merging this is fine.
But please comment (or open a new issue) if you find something else or have a comment.

mallet:
Thank you very much for implementing this @dakra ! Much appreciated.
I will use this version, obviously, and let you know if I find something else.

PR #665, opened 2026-09-01
Expose live terminal send boundary

miguelmartin75:
## Summary

- add the public `ghostel-buffer-live-p` predicate for input-ready Ghostel buffers
- make `ghostel-paste-string` and `ghostel-send-key` signal `user-error` when a PTY is already dead or exits around a write
- document the liveness and send-failure contracts
- cover live, exited, and exit-racing sends at the public Elisp boundary

## Testing

- focused pure-Elisp key suite: 64/64 passed
- complete native test target: passed with 8 environment-dependent skips
- byte compilation, Checkdoc, docquote lint, `check-parens`, and `git diff --check` passed

### DIFF
--- a/README.org
+++ b/README.org
@@ -2011,21 +2011,25 @@ to show it in every buffer prompt (=C-x b=, =consult-buffer=, and ghostel's own)
 :end:
 #+findex: ghostel-send-string
 #+findex: ghostel-send-key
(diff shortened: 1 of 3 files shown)


miguelmartin75:
Sorry, this is slop. Closing.

PR #669, opened 2026-09-01
Skip clean rows in incremental redraws

ivanrg99:
Start rendering at the first dirty row instead of traversing the viewport from the top. This avoids walking unchanged rows in tall terminals while preserving full redraw behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ All notable changes to this project will be documented in this file.
   `bottom` moves it all above the grid, `top` (the default) keeps it
   below.  Graphical frames on Emacs 29+ only.
 
+### Changed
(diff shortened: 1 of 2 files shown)


emil-e:
I did some profiling and benchmarking here and I cannot really reproduce any meaningful improvement in performance due to this. Is there any specific bottleneck you are targetting here?

ivanrg99:
You are correct, this optimization was mostly to reduce the cost that https://github.com/dakra/ghostel/pull/668 would introduce. Maybe I should move these changes to the other PR ? I also realized that when I split it I did not really leave a good explanation of what it does, my apologies. I won't be having much free time in some time, but I'll get back to it as soon as I can. You may feel free to close it in the meantime if that is not ok, my apologies again

PR #686, opened 2026-09-14
Report cell pixel size in the PTY window size and enable all kitty mediums

### LINKED ISSUE #675: Kitty graphics protocol image corruption (likely inherited from vendored ghostty commit, see ghostty-org/ghostty#12901)
### Description

Images rendered via the Kitty graphics protocol inside a ghostel buffer come out corrupted — mostly blank canvas with scattered, garbled colored blocks — while the same image renders correctly in a raw Ghostty window on the same system. This affects at least PDF-preview-via-image tooling (specifically broot's `preview_transformers`, using `mutool` to rasterize PDFs to PNG before Kitty-protocol transfer).

### Environment

- ghostel version: `20260902.1753` (commit `14c399002f47405a623604170314e27d260efb30`)
- Native module version: `0.53.0`
- Emacs: 30.2, terminal (not GUI), native-comp
- OS: NixOS, Hyprland, Wayland
- System Ghostty package: 1.3.1

### Reproduction

1. Open a ghostel buffer (`M-x ghostel`)
2. Run `broot` with a `preview_transformers` config converting PDFs to PNG via `mutool draw -r 300`
3. Preview any PDF file
4. Resulting image is corrupted (blank canvas + scattered colored artifact blocks); expected: correctly rendered rasterized PDF page

### What I've ruled out

- **Not a resolution/DPI issue**: source PNG confirmed correct at 3508×2481px via `mutool`.
- **Not a broot bug**: identical preview pipeline renders correctly in a raw Ghostty window (not through ghostel) on the same system. 

- image diffs:
  Here is an image of broot preview running on a raw ghostty window:

Image
   Here is an image running in ghostel buffer (inside emacs)

Image

- **Not a `TERM`/`TERMINAL` detection issue**: `$TERMINAL` and `$TERM_PROGRAM` are both correctly reported as `ghostty` from inside ghostel.
- **Confirmed via `ghostel-debug-info`**: ghostel reports `TERM_PROGRAM_VERSION=1.3.2` — this appears to trace to the vendored Ghostty commit pinned in `build.zig.zon` (`01825411ab2720e47e6902e9464e805bc6a062a1`, hash-tagged `ghostty-1.3.2-dev-...`), dated 2026-04-13.

### Likely root cause

This looks like the same bug reported upstream in [ghostty-org/ghostty#12901](https://github.com/ghostty-org/ghostty/discussions/12901): `error.StreamTooLong` being thrown as fatal in `src/terminal/kitty/graphics_image.zig`'s `appendRemaining` call when reading Kitty-protocol image data via the temp-file (`t=t`) transmission medium. I independently reproduced and confirmed this exact fix resolves the corruption in a raw Ghostty 1.3.1 build (patched via local overlay, verified running via store-path check).

Since ghostel's vendored commit (`0182541`) predates that discussion's report by about 7 weeks with no evidence of the fix landing on `main` since, it's plausible ghostel is simply inheriting the still-unfixed upstream bug.

### Request

- Could the vendored ghostty commit in `build.zig.zon` be bumped once a fix lands upstream?
- Alternatively, if this turns out to be a distinct bug in ghostel's own `ghostel-kitty` handling rather than the vendored engine, happy to help narrow it down further — I can provide `ghostel-debug-start` logs from a live repro if useful.

dakra:
Fixes #675.

The reporter's "corrupted" broot preview is broot's half-block text fallback, which it uses when it cannot determine the cell pixel size. Two ghostel-side causes:

- The native PTY backend set `ws_xpixel`/`ws_ypixel` to 0 in `TIOCSWINSZ`. broot (≤ 1.58, the nixpkgs stable version) reads `TIOCGWINSZ` only and disables kitty graphics on zero pixel fields; ranger reads it too.
- `ghostel-kitty-graphics-mediums` defaulted to nil, so broot's default temp-file / file transmissions were refused with `EINVAL: unsupported medium`. broot sends `q=2` and never sees the refusal.

Changes:

- A `WinSize` struct (cols, rows, xpixel, ypixel) flows from `GhostelTerm` through `NativeProcess` to the PTY backends, filled from libghostty's `width_px`/`height_px`. `GhostelTerm` tracks the last size sent to the native PTY and resends after a redraw when the grid or the pixel geometry changed, so DPI and cell-scale changes propagate too. The Emacs-owned PTY path is unchanged (`set-process-window-size` cannot carry pixels). ConPTY ignores the pixel fields.
- `ghostel-kitty-graphics-mediums` defaults to `(file temp-file shared-mem)`, as in Ghostty and kitty. nil keeps the inline-only behavior.
- Native ERT test: a python child reads `TIOCGWINSZ` and sees the spawn-time seed and a 9×23 resize as `720×552` px.

Verified in a live GUI Emacs: a child sees `WINSZ (33, 100, 1100, 726)`, consistent with the CSI 16 t reply, and a broot-style `t=t` transmit returns `OK` and renders.

Not covered: ranger's capability probe sends `S=3` and hits the libghostty exact-size read bug (upstream ghostty-org/ghostty#12901, fixed in ghostty abd77067d), so ranger still needs a libghostty bump. ranger also requires `TERM` to contain "kitty".


src/backend_types.zig
```diff
@@ -7,6 +7,14 @@ pub const ProcessParams = struct {
     cwd: ?[:0]const u8 = null,
 };
 
+/// PTY window size.  ConPTY has no pixel geometry and ignores the pixel fields.
```
emil-e: Comment can be removed.

emil-e (approved):
LGTM

PR #690, opened 2026-09-15
Prevent terminal query replies from blocking Emacs

### LINKED ISSUE #689: Native PTY: terminal query reply blocks the reader thread with the terminal lock held, freezing Emacs
## Summary

With `ghostel-use-native-pty t`, a program that emits terminal queries faster than it
drains its own stdin deadlocks Emacs permanently.

The VT reply is written back into the pty from *inside* the parse loop, while the
terminal mutex is held, and with a `null` cancellation token — so it blocks in
`poll(POLL.OUT)` forever once the pty input queue fills. The Emacs main thread then
wedges on its next `ghostel--redraw`.

Emacs stays alive but is completely unresponsive: no redisplay, no input, and
`emacsclient` hangs. Attaching with `lldb` hangs too, which makes it awkward to debug
in place.

## Versions

- ghostel **0.53.0** (native module 0.53.0) — code paths below are unchanged at
  **v0.54.0**, so this is still present on latest
- ghostty dependency 1.3.2-dev (`ab0b9da`, per `build.zig.zon`)
- GNU Emacs 31.1, macOS NS build
- macOS 26.6.2 (25G83), arm64

## Reproduction

Wedges a clean `emacs -Q` in about ten seconds.

`prog.py` — stands in for a full-screen TUI:

```python
import sys, time, tty
tty.setraw(sys.stdin.fileno())      # raw mode, like a real TUI
w = sys.stdout.write
for _ in range(20000):
    w('\x1b[6n')                    # DSR - ghostel must reply into the pty
    w('x' * 200 + '\r\n')           # flood, so the reader thread stays in the parse
sys.stdout.flush()
time.sleep(600)                     # never read stdin
```

`init.el`:

```elisp
(setq ghostel-use-native-pty t)
(require 'ghostel)
(setq ghostel-shell (list "/usr/bin/python3" "/path/to/prog.py"))
(run-at-time 2 nil #'ghostel)
```

```
emacs -Q -L /path/to/ghostel -l /path/to/init.el
```

**Both halves are needed.** 500 queries with no flooding, tty left in canonical mode,
does *not* reproduce — the queue drains and everything recovers. The reader thread has
to still be inside the parse, holding the lock, at the moment the write blocks.

### How this was originally hit

The program above is a minimization. The original sighting was simply:

1. open a ghostel buffer
2. `ssh <remote-host>`
3. run `herdr` (a full-screen terminal multiplexer) on the remote

which hangs reliably.

What matters is **which process holds the far end of the pty**, not that anything is
remote:

- With `ssh` as ghostel's direct pty child, it wedges. ssh is busy relaying a burst of
  full-screen TUI output and does not drain its own stdin promptly, so the replies
  ghostel writes back accumulate until the input queue is full.
- With `herdr` itself as the pty child — connected to a *remote* server, so ssh was still
  in the process tree — several minutes of driving the TUI, including 200 scripted window
  resizes, did **not** reproduce. herdr drains its stdin regardless of where its panes
  live.

So the ingredient is an intermediary that is slow to read its own stdin, and `ssh` is
just the most common example of one. Network latency does not appear to be the relevant
factor, and "local vs remote" is not the right axis.

## Observed

Once wedged:

- `emacsclient -e '(+ 1 1)'` never returns.
- `ioctl(FIONREAD)` on the pty replica is pinned at **1022** bytes (the 1024-byte
  limit) and never moves.
- `sample <emacs-pid>` shows one module thread several frames deep ending in `poll` —
  the parked write-back — and the main thread blocked wherever it next needed the
  terminal. In the original sighting that was `__ulock_wait2` beneath
  `ghostel--redraw-now`; in a run driven by scripted `ghostel-send-string` calls it was
  the send path instead (`nanosleep` / `__semwait_signal` inside the module). The
  invariant is the parked writer plus the pinned queue, not any one victim stack.

Reading the replica dry recovers it — open `/dev/ttysNNN` with
`O_RDWR | O_NOCTTY | O_NONBLOCK` and read until `EAGAIN`. The master turns writable, the
lock releases, and Emacs resumes with every buffer intact. One such recovery drained
~140 KB, so the blocked writer had a substantial backlog queued behind it.

**Killing the child does not help.** `SIGKILL` on the program leaves it a zombie and
Emacs still wedged: macOS does not wake a `POLL.OUT` wait on the pty master when the
last replica closes, so the thread stays parked. Verified directly against the repro
above.

## Cause

All on the reader thread (line numbers at v0.54.0):

1. `src/NativeProcess.zig:49` — `LockedStream.nextSlice` takes `self.process.owner.lock()`
   and holds it across `stream.nextSlice`.
2. `src/handler.zig:38` — `self.inner.effects.write_pty = &writePtyCallback`.
3. `src/handler.zig:96` — `writePtyCallback` fires from inside that parse and calls
   `effects.ptyWriteFromTerminal(data)`.
4. `src/NativeProcess.zig:129` — `backend.write(data[offset..], null)` — **null**
   cancellation token.
5. `src/PosixPtyProcess.zig:295-301` — on `EAGAIN`, the timeout is `-1` precisely when
   `cancellation == null`:

```zig
const timeout: i32 = if (cancellation) |token|
    @intCast(@min(
        token.poll_interval.toMilliseconds(),
        @as(u32, @intCast(std.math.maxInt(i32))),
    ))
else
    -1;
const ready = pollWithRetry(&pollfds, timeout);
```

So `pollWithRetry` waits forever on `POLL.OUT` with the terminal mutex still held from
step 1, and any `ghostel--redraw` on the main thread blocks behind it.

For what it's worth this looks like ghostel's side of the boundary rather than
libghostty's: `write_pty` is a documented embedding hook, and
`include/ghostty/vt/terminal.h` states *"Callbacks are invoked synchronously during VT
writes"* — so not blocking inside the callback is the embedder's job.

## Possible fix

The plumbing is already there:

- The cancellation parameter is honoured at every other call site; this one passes
  `null`. Passing a token (or any bounded timeout) would turn the permanent hang into a
  retry loop that can observe `self.quit`.
- `PosixPtyProcess.write` already polls `wake_pipe[0]` alongside `POLL.OUT`, so an
  interrupt path exists. The comment at `NativeProcess.zig:118-120` notes the handoff
  mutex is skipped deliberately so shutdown can signal a blocked write — but nothing
  signals it while the session is live.

Dropping or bounding the reply when the input queue is full may also be defensible: a
program that is not reading its stdin was never going to consume a DSR reply anyway.

emil-e:
## Summary

- make terminal-generated PTY replies abort immediately when the child input queue is full
- drop blocked replies instead of waiting with the terminal lock held
- retain serialization with cancellable Emacs-originated input

Fixes #689.

## Verification

- `zig build --prefix .`
- `zig build test`
- native-process ERT tests (5/5)
- reproduction-style smoke test with 20,000 DSR queries


srijan:
Tested this locally. It fixes the freeze, and I found one residual issue worth a follow-up.

## Setup

Built both `b46898e` (PR head) and `f1b03e5` (its base) from source with zig 0.16.0, `-Doptimize=ReleaseFast`, and ran each in a throwaway `emacs -Q` with `ghostel-module-directory` pointed at the respective build and `load-path` at the matching `lisp/`, so the elisp and the module never skew. macOS 25.6, Emacs 31.1, `ghostel-use-native-pty t`.

Wedge detection is an `emacsclient --eval` probe against a `server-start` in the test Emacs, with `timeout`; exit 124 means Emacs never got back to the command loop.

## Results

The original reproduction from #689 (20,000 DSR queries interleaved with output, child never reads stdin):

| | base `f1b03e5` | PR `b46898e` |
|---|---|---|
| `emacsclient` probe | exit 124 (wedged) | exit 0 |
| threads parked in `__ulock_wait` | 2 | 0 |

The real-world shape that originally hit this — `ssh` as ghostel's direct pty child, with a full-screen TUI started on the far end:

| | base `f1b03e5` | PR `b46898e` |
|---|---|---|
| probes over 60s | 5/5 timed out | 6/6 answered |
| threads parked in `__ulock_wait` | 2 | 0 |

Under the PR the remote TUI rendered correctly and Emacs stayed responsive throughout.

## Regression checks

Both pass on the PR, so the zero-timeout give-up does not fire for programs that actually read their input:

- Plain DSR handshake, a program that queries and reads the reply: 5/5 replies received, identical on both builds.
- Busy-but-healthy TUI: 2,000 DSR queries interleaved with 400-byte output bursts, child draining stdin promptly. 2,000/2,000 replies received on both builds, zero dropped.

`zig build test` passes.

## One residual: the drop is not all-or-nothing

`ptyWriteFromTerminal` advances `offset` across partial writes, so when the input queue fills *mid-reply* it abandons a half-written escape sequence in the child's input. The old code never did this, because it blocked until the whole reply landed — which is of course the freeze being fixed here.

Repro: fill the queue from a child that is not reading, then drain it and look at what is actually sitting there.

```python
import sys, os, re, time, tty
fd = sys.stdin.fileno()
tty.setraw(fd)
w = sys.stdout.write
for _ in range(4000):                 # stuff the queue while never reading
    w('\x1b[6n'); w('z' * 100 + '\r\n')
sys.stdout.flush()
time.sleep(3)
os.set_blocking(fd, False)
buf = b''
end = time.time() + 5
while time.time() < end:
    try: buf += os.read(fd, 65536)
    except Exception: time.sleep(0.05)
whole = re.findall(rb'\x1b\[\d+;\d+R', buf)
leftover = re.sub(rb'\x1b\[\d+;\d+R', b'', buf)
print('bytes=%d well_formed=%d non_reply_bytes=%d\nleftover=%r'
      % (len(buf), len(whole), len(leftover), leftover[:64]))
```

On the PR build:

```
bytes=1022 well_formed=146 non_reply_bytes=4
leftover=b'\x1b[34'
```

That trailing `ESC [ 3 4` is a DSR reply cut in half at the 1024-byte queue boundary. A TUI that recovers and resumes reading will treat it as the start of a sequence and swallow the next keystrokes into it.

Suggested narrowing: give up only at `offset == 0`, so a terminal reply is either written whole or not written at all. Partial delivery of a reply is never useful to the child.

None of this is a reason to hold the PR — a mangled escape sequence is a far better failure than a frozen editor. Happy to test a follow-up.

emil-e:
Thanks for testing. On the residual, I would decline the narrowing since that reintroduces the original bug. For a pathological child, this would again make it possible for it to block forever. I will merge this as is.

PR #709, opened 2026-09-19
Render no-break spaces with the default face

### LINKED ISSUE #708: No-break spaces (U+00A0) use a visible face
Steps to reproduce:

```sh
printf "\xc2\xa0\n"
```

aintbck:
No-break spaces inherit Emacs's `nobreak-space` face. Themes such as Modus render them as colored marks in terminal output.

Remap the face locally so no-break spaces render like normal terminal spaces.

Closes: #708


dakra:
Thanks for the PR.
I fixed the issue in #710 slightly differently.

